PR #1, opened 2023-05-26
Monorepo setup

konraddysput:
# Why

This diff adds monorepo basic setup to the repository.


packages/core/src/BacktraceCoreClient.ts
```diff
@@ -0,0 +1,8 @@
+import { BacktraceAttachment } from './model/report/BacktraceAttachment';
+import { BacktraceReport } from './model/report/BacktraceReport';
+
+export class BacktraceCoreClient {
+    public async send(report: BacktraceReport, attributes: Record<string, any>, attachments: BacktraceAttachment) {
```
perf2711, line 5: why `attributes` and `attachments` are not part of `BacktraceReport`?

packages/core/src/model/configuration/BacktraceConfiguration.ts
```diff
@@ -0,0 +1,51 @@
+import { BacktraceAttachment } from '../report/BacktraceAttachment';
+import { BacktraceDatabaseConfiguration } from './BacktraceDatabaseConfiguration';
+
+// Submission information
+export type BacktraceBasicSubmissionInformation = {
+    /**
+     * The server address (submission URL) is required to submit exceptions from your Unity project to your Backtrace instance.
```
perf2711, line 7: >Unity

I think you missed this part

packages/core/src/BacktraceCoreClientBuilder.ts
```diff
@@ -0,0 +1,8 @@
+import { BacktraceCoreClient } from './BacktraceCoreClient';
+import { BacktraceConfiguration } from './model/configuration/BacktraceConfiguration';
+
+export class BacktraceCoreClientBuilder {
+    public static build(options?: BacktraceConfiguration): BacktraceCoreClient {
+        throw new Error('Not implemented');
+    }
+}
```
perf2711, line 8: Will we ever build core client, if other clients will derive from it?

packages/core/src/BacktraceCoreClient.ts
konraddysput: Done. Thx

packages/core/src/model/configuration/BacktraceConfiguration.ts
konraddysput: Done Thx

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -0,0 +1,7 @@
+import { BacktraceAttachment } from './BacktraceAttachment';
+
+export interface BacktraceReport {
+    attributes(): Record<string, unknown>;
+    attachments(): BacktraceAttachment[];
+    toDate(): Record<string, unknown>;
```
perf2711, line 6: what is `toDate` exactly? Is there a reason that these are methods, and not `readonly` fields?

konraddysput: those are methods and generate something  - toDate convert report to the data object that we can send to Backtrace. Those information will more clean soon

konraddysput: Ups, I see - typo

packages/node/package.json
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "@backtrace/node",
+    "version": "0.0.1",
+    "description": "Backtrace-JavaScript Node.JS integration",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "test": "NODE_ENV=test jest"
```
perf2711, line 8: we are missing `build`, `rebuild` etc. scripts here

packages/browser/package.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "@backtrace/browser",
+    "version": "0.0.1",
+    "description": "Backtrace-JavaScript web browser integration",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "test": "NODE_ENV=test jest"
```
perf2711, line 8: we are missing `build`, `rebuild` etc. scripts here

packages/sdk-core/package.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "@backtrace/sdk-core",
+    "version": "0.0.1",
+    "description": "Backtrace-JavaScript SDK core library",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "test": "NODE_ENV=test jest"
```
perf2711, line 8: we are missing build, rebuild etc. scripts here

packages/sdk-core/src/model/configuration/BacktraceConfiguration.ts
```diff
@@ -0,0 +1,51 @@
...
+     */
+    attachments?: BacktraceAttachment[];
+
+    /**
+     * Metrics such as crash free users and crash free sessions
+     */
+    metrics?: BacktraceMetricsSupport;
+};
+
+export type BacktraceConfiguration = BasicBacktraceClientSettings & BacktraceDatabaseConfiguration;
```
perf2711, line 51: why are we not using `interface` here in these options (not the union ones)?

packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
```diff
@@ -0,0 +1,71 @@
+export enum DeduplicationStrategy {
+    None = 0,
+    Callstack = 1 << 0,
+    Classifier = 1 << 1,
+    Message = 1 << 2,
+    All = ~(~0 << 4),
+}
+export type EnabledBacktraceConfiguration = {
```
perf2711, line 8: why not `interface`?

packages/sdk-core/src/model/report/BacktraceAttachment.ts
```diff
@@ -0,0 +1,13 @@
+export type BacktraceInMemoryAttachment = {
```
perf2711, line 1: why are we not using `interface` here in these options?

packages/node/package.json
konraddysput: done

packages/browser/package.json
konraddysput: done

packages/sdk-core/package.json
konraddysput: done

packages/sdk-core/src/model/configuration/BacktraceConfiguration.ts
konraddysput: why should we? Do we want to inherit from them? Is there any better reason to use interface than type?

perf2711: it seems more suitable to use an interface to define an object contract. I would use `type` to define a type alias, like  `type BacktraceConfiguration = BasicBacktraceClientSettings & BacktraceDatabaseConfiguration;`, or `type Key = string | number`.

konraddysput: Pushed adjusted code base on our conversation

packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
konraddysput: Pushed adjusted code base on our conversation

packages/sdk-core/src/model/report/BacktraceAttachment.ts
konraddysput: Pushed adjusted code base on our conversation

perf2711 (approved):

PR #5, opened 2023-05-30
Backtrace report generation

konraddysput:
# Why

This diff shows how to construct basic Backtrace report and how to send it.

### DIFF
--- a/packages/sdk-core/src/BacktraceCoreClient.ts
+++ b/packages/sdk-core/src/BacktraceCoreClient.ts
@@ -1,7 +1,45 @@
+import { BacktraceAttachment } from './model/report/BacktraceAttachment';
 import { BacktraceReport } from './model/report/BacktraceReport';
 
 export class BacktraceCoreClient {
-    public async send(report: BacktraceReport) {
-        throw new Error('Not implemented');
+    /**
+     * Asynchronously sends error data to Backtrace.
+     * @param error Backtrace Report or error or message
+     * @param attributes Report attributes
+     * @param attachments Report attachments
+     */
+    public async send(
+        error: Error,
+        attributes?: Record<string, unknown>,
+        attachments?: BacktraceAttachment[],
+    ): Promise<void>;
+    /**
+     * Asynchronously sends a message report to Backtrace
+     * @param message Report message
+     * @param attributes Report attributes
+     * @param attachments Report attachments
+     */
+    public async send(
+        message: string,
+        attributes?: Record<string, unknown>,
+        attachments?: BacktraceAttachment[],
+    ): Promise<void>;
+    /**
+     * Asynchronously sends error data to Backtrace
+     * @param report Backtrace Report
+     */
+
+    public async send(report: BacktraceReport): Promise<void>;
+    public async send(
+        data: BacktraceReport | Error | string,
+        attributes: Record<string, unknown> = {},
+        attachments: BacktraceAttachment[] = [],
+    ): Promise<void> {
+        let report = this.isReport(data) ? data : new BacktraceReport(data, attributes, attachments);
+        console.log(report.message);
+    }
+
+    private isReport(data: BacktraceReport | Error | string): data is BacktraceReport {
+        return !(data instanceof Error || typeof data === 'string');
     }
 }
--- a/packages/sdk-core/src/model/report/BacktraceReport.ts
+++ b/packages/sdk-core/src/model/report/BacktraceReport.ts
@@ -1,7 +1,61 @@
 import { BacktraceAttachment } from './BacktraceAttachment';
 
-export interface BacktraceReport {
-    readonly attributes: Record<string, unknown>;
-    readonly attachments: BacktraceAttachment[];
-    toData(): Record<string, unknown>;
+export class BacktraceReport {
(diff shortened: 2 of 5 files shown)


packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -1,7 +1,45 @@
...
+        data: BacktraceReport | Error | string,
+        attributes: Record<string, unknown> = {},
+        attachments: BacktraceAttachment[] = [],
+    ): Promise<void> {
+        const report = this.isReport(data) ? data : new BacktraceReport(data, attributes, attachments);
+        console.log(report.message);
+    }
+
+    private isReport(data: BacktraceReport | Error | string): data is BacktraceReport {
+        return !(data instanceof Error || typeof data === 'string');
```
perf2711, line 43: Why not do something like `data instanceof BacktraceReport`?

packages/sdk-core/src/model/report/BacktraceErrorType.ts
```diff
@@ -0,0 +1 @@
+export type BacktraceErrorType = 'Message' | 'Exception' | 'Uhnandled exception' | 'OOMException' | 'Hang';
```
perf2711, line 1: if we're having `Unhandled exception` with a whitespace, maybe we should also have `OOMException`?

konraddysput: @perf2711 we follow everything that we have in our libraries - I believe we can make an excuse for OOMException but let's not do that now. I will need to verify all our libraries to verify that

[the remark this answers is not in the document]
konraddysput: ups, thanks!

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -1,7 +1,52 @@
...
+        } else {
+            this.message = data;
+            this.stackTrace = new Error().stack ?? '';
+            errorType = 'Message';
+        }
+
+        if (!this.attributes['error.type']) {
+            this.attributes['error.type'] = errorType;
+        }
+    }
```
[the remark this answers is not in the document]
konraddysput: the use case for this for our clients is to write 

try {
} catch(err) {
if(something) {
const report =new BacktraceReport(err);
client.send(report);
}
} 

if we generate additional static methods like fromError it might generate a lot of confusion for our users

packages/sdk-core/src/BacktraceCoreClient.ts
konraddysput: I didn't know it works everywhere. It didn't work in the past, Let me try.

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -1,7 +1,58 @@
...
+
+    /**
+     * Report inner errors
+     */
+    public readonly innerReport: unknown[] = [];
+
+    /**
+     * Number of frames generated by the Backtrace SDK
+     */
+    public libraryGeneratedFrames = 0;
```
perf2711, line 31: maybe let's name this just `skipFrames`?

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -1,7 +1,49 @@
...
+
+    public async send(report: BacktraceReport): Promise<void>;
+    public async send(
+        data: BacktraceReport | Error | string,
+        attributes: Record<string, unknown> = {},
+        attachments: BacktraceAttachment[] = [],
+    ): Promise<void> {
+        const report = this.isReport(data) ? data : new BacktraceReport(data, attributes, attachments);
+        if (typeof data === 'string') {
+            report.libraryGeneratedFrames++;
```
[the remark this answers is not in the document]
konraddysput: not really - if you have an error, you shouldn't increase the frame.

packages/sdk-core/src/model/report/BacktraceReport.ts
konraddysput: sure!

packages/sdk-core/src/BacktraceCoreClient.ts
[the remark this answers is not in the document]
konraddysput: I've added Backtrace report options to manage it better. What do you think about it?

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -1,7 +1,60 @@
...
+    public readonly message: string;
+
+    /**
+     * Report inner errors
+     */
+    public readonly innerReport: unknown[] = [];
+
+    /**
+     * Number of frames generated by the Backtrace SDK
+     */
```
perf2711, line 30: The doc is invalid.

perf2711 (approved):

packages/sdk-core/src/model/report/BacktraceReport.ts
konraddysput: Adjusted the comment based on our conversation

PR #6, opened 2023-06-06
Stack trace parser

konraddysput:
# Why

This pull request adds a stack trace parser to the SDK core library.  Stack Trace parsers generate stack frames needed to generate a Backtrace report

More details:
- In the past, we used regexes to convert stack frames into data objects. This wasn't a good approach, since our regexes needed updates that we didn't do. Because of that, we detected some issues on Firefox etc
- regexes are excellent and fast! However, the problem starts when no one knows how to modify it. This code uses simple string manipulations to parse the stack frame. If we keep it simple and easy to understand, we will have a quick and easy way to iterate on this if we detect something
- but Konrad, why we do not parser per specific SDK? Actually, everything else uses v8 - except the web where we can find different stack traces for Safari and Firefox. With this knowledge, to avoid potential headaches and additional work for other SDK maintainers, I decided to put it in the common logic. 
- What is an anonymous frame? As you know, we need to care about fingerprints that we're generating. Depending on this we can generate different fingerprints for the same stack trace generated by different web browsers. To avoid this, anonymous frames are converted into the same anonymous path


packages/sdk-core/src/model/data/BacktraceStackTrace.ts
```diff
@@ -0,0 +1,12 @@
+export interface BacktraceStackFrame {
+    funcName: string;
+    line?: number;
+    column?: number;
+    sourceCode?: string;
+    library: string;
```
perf2711, line 6: By `library`, do you mean `filename`?

packages/sdk-core/src/model/data/BacktraceStackTrace.ts
```diff
@@ -0,0 +1,12 @@
...
+}
+export interface BacktraceStackTrace {
+    name: string;
+    fault: boolean;
```
perf2711, line 10: What does `fault` mean? Can you add descriptions?

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
```diff
@@ -0,0 +1,161 @@
+import { BacktraceStackFrame } from '../../model/data/BacktraceStackTrace';
+import { JavaScriptEngine } from '../../model/data/JavaScriptEngine';
+
+export class StackTraceConverter {
+    public readonly UNKNOWN_FRAME = 'unknown';
+    private readonly ANONYMOUS_FRAME: Record<JavaScriptEngine, string> = {
+        ['firefox']: 'debugger eval code',
+        ['v8']: '<anonymous>',
+        ['safari']: '[native code]',
```
perf2711, line 9: How did you get those values? I only get them in Firefox and Chrome when using the console, but when I serve a file, those are not printed. I've never seen `[native code]` of Safari even once.

### Firefox 
image

### Chrome
image

### Safari
image


### Testing code

```
function lambdaAnonymousTest() {
    (() =>
        (() =>
            (() => {
                throw new Error('lambdaAnonymousTest');
            })())())();
}

function anonymousTest() {
    (function () {
        (function () {
            (function () {
                throw new Error('anonymousTest');
            })();
        })();
    })();
}

function namedLambdaAnonymousTest() {
    const throwing = () => {
        throw new Error('namedLambdaAnonymousTest');
    };
    const anon2 = () => throwing();
    const anon1 = () => anon2();
    anon1();
}

function namedTest() {
    function throwing() {
        throw new Error('namedTest');
    }

    function anon2() {
        throwing();
    }

    function anon1() {
        anon2();
    }

    anon1();
}

function test(fun) {
    try {
        fun();
    } catch (err) {
        console.log(err.message, err.stack);
    }
}

test(lambdaAnonymousTest);
test(anonymousTest);
test(namedLambdaAnonymousTest);
test(namedTest);
```

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
```diff
@@ -0,0 +1,161 @@
...
+        const functionSeparator = '@';
+        const functionSeparatorIndex = stackFrame.indexOf(functionSeparator);
+        if (functionSeparatorIndex === -1) {
+            return {
+                funcName: stackFrame,
+                library: this.UNKNOWN_FRAME,
+            };
+        }
+
+        const functionName = stackFrame.substring(0, functionSeparatorIndex) ?? this.UNKNOWN_FRAME;
```
perf2711, line 62: From what I've seen in Safari, if there's nothing before `@`, it means that the function was anonymous.

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
```diff
@@ -0,0 +1,161 @@
...
+        const frameSeparator = 'at ';
+        if (!stackFrame.startsWith(frameSeparator)) {
+            return {
+                funcName: stackFrame,
+                library: this.UNKNOWN_FRAME,
+            };
+        }
+
+        stackFrame = stackFrame.substring(stackFrame.indexOf(frameSeparator) + frameSeparator.length);
+        const sourceCodeSeparator = ' (';
```
perf2711, line 142: The separator does not appear with anonymous functions:

```
Error: lambdaAnonymousTest
    at http://localhost:8080/index.js:5:23 <-- anonymous
    at http://localhost:8080/index.js:6:15 <-- anonymous
    at http://localhost:8080/index.js:6:18 <-- anonymous
    at lambdaAnonymousTest (http://localhost:8080/index.js:6:21)
    at test (http://localhost:8080/index.js:52:9)
    at http://localhost:8080/index.js:58:1 <-- anonymous
```

packages/sdk-core/tests/converters/stackTraceConverterTests.spec.ts
```diff
@@ -0,0 +1,152 @@
+import { StackTraceConverter } from '../../src/modules/converter/StackTraceConverter';
```
perf2711, line 1: You should add tests with anonymous functions.

If we're trying to build a robust solution, I would start with supporting all cases listed here:
https://github.com/stacktracejs/error-stack-parser/blob/master/spec/fixtures/captured-errors.js
(the code is MIT, so I think you could just copy the sample stacks)

perf2711 (changes requested):

packages/sdk-core/src/model/data/BacktraceStackTrace.ts
konraddysput: api.backtrace.io

packages/sdk-core/src/model/data/BacktraceStackTrace.ts
konraddysput: This is the definition from the api.backtrace.io and represents the thread object we want to send. Backtrace expects filename under the library unless something has changed.

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
```diff
@@ -0,0 +1,161 @@
...
+        // remove error header from stack trace - if the error header exists
+        if (stackFrames[0].indexOf(errorHeader[0]) !== -1) {
+            stackFrames = stackFrames.slice(errorHeader.length);
+        }
+
+        for (const stackFrame of stackFrames) {
+            const normalizedStackFrame = stackFrame.trim();
+            const frame = this.parseFrame(normalizedStackFrame);
+            if (this.ANONYMOUS_FRAME[this._lang] === frame.library) {
+                frame.library = this.UNKNOWN_FRAME;
```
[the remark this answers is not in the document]
adamcronin42: yeah, we could mark it as "anonymous" instead of "unknown", or leave as is

packages/browser/src/engineDetector.ts
```diff
@@ -0,0 +1,12 @@
+import { JavaScriptEngine } from '@backtrace/sdk-core/src/model/data/JavaScriptEngine';
+
+export function getEngine(): JavaScriptEngine {
+    if (navigator.userAgent.includes('Firefox')) {
+        return 'firefox';
+    }
+    if (navigator.userAgent.includes('Safari')) {
```
adamcronin42, line 7: We may want to find a more concrete way of doing this. When I run `navigator.userAgent` in my Chrome console it gives this output: 
`> navigator.userAgent
'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'` which would return true here.

packages/sdk-core/src/model/data/BacktraceStackTrace.ts
konraddysput: done.

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
konraddysput: done.

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
konraddysput: done.

packages/sdk-core/src/modules/converter/StackTraceConverter.ts
konraddysput: done

packages/browser/src/engineDetector.ts
konraddysput: Ok, once again the tactic has changed - this time based on some materials on the Internet I'm verifying If the web browser is Chromium base and if the user agent contains information about Safari. https://www.geeksforgeeks.org/how-to-detect-the-user-browser-safari-chrome-ie-firefox-and-opera-using-javascript/

adamcronin42 (approved):

packages/sdk-core/src/modules/converter/BacktraceStackTraceConverter.ts
```diff
@@ -0,0 +1,8 @@
+import { BacktraceStackFrame } from '../../model/data/BacktraceStackTrace';
+import { JavaScriptEngine } from '../../model/data/JavaScriptEngine';
+import { BacktraceReport } from '../../model/report/BacktraceReport';
+
+export interface BacktraceStackTraceConverter {
+    get engine(): JavaScriptEngine;
```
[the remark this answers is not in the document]
konraddysput: We will inject it via the factory - I still believe the getter is useful to have, so this is why we have it. We will always have one converter, so the canConvert method won't be useful.

perf2711 (approved):

PR #7, opened 2023-06-06
Data submission API

konraddysput:
# Why

We know our library needs to post data in at least two reasons: 
* submit data to Backtrace (minidump, json, plcrashreporter, but lets talk in this diff only about json submission)
* send metrics information

As you know the package size matters! In the past when we were working on the backtrace-js, we wanted to avoid adding any unnecessary dependency that might increase the build size on web. In addition to that, Node doesn't have XmlHttpRequest API that is available in web browser. Also, in node, we don't care as much about the size as we care about the web browser. 

With the long/short introduction, this diff allows us to submit all kinds of data to Backtrace from the SDK via our new submission layer. By doing that we should be able to see errors in web browser/node in any format and also we should be able to send metrics.


packages/node/src/BacktraceNodeRequestHandler.ts
```diff
@@ -0,0 +1,107 @@
+import {
+    BacktraceAttachment,
+    BacktraceReportSubmissionResult,
+    BacktraceRequestHandler,
+    BacktraceSubmissionResponse,
+    DEFAULT_TIMEOUT,
+} from '@backtrace/sdk-core';
+import { BacktraceData } from '@backtrace/sdk-core/src/model/data/BacktraceData';
+import axios from 'axios';
```
perf2711, line 9: Can we _not_ use `axios`? Let's keep the dependency list as short as possible.

Instead, let's use `http` and `https`, and on Node 18, we can use `fetch`.

packages/browser/src/BacktraceBrowserRequestHandler.ts
```diff
@@ -0,0 +1,92 @@
...
+    ): Promise<BacktraceReportSubmissionResult<T>> {
+        const formData = this.createFormData(JSON.stringify(data), attachments);
+        return this.post(submissionUrl, formData);
+    }
+
+    public async post<T>(
+        submissionUrl: string,
+        payload: string | FormData,
+    ): Promise<BacktraceReportSubmissionResult<T>> {
+        const xmlHttpRequest = new XMLHttpRequest();
```
perf2711, line 35: Is there a reason why you're using XHR instead of fetch? All modern browsers seem to support it.

https://caniuse.com/fetch

konraddysput: This is the copy paste code from Backtrace-js. We can use fetch

packages/node/src/BacktraceNodeRequestHandler.ts
konraddysput: instead of supporting http/https and fetch on node 18 it's worth sticking to one implementation. In addition to that on the server side, we don't care much about the package size. we used http/https in our migration script and until now I'm not sure how we can handle specific events generated by them. http/https might introduce additional pain and won't give us too much. 

I agree we should use less packages than more, but is there a real benefit for not using axios if we use it everywhere?

packages/sdk-core/src/model/http/model/BacktraceSubmissionResult.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    public static OnInvalidToken<T>(): BacktraceReportSubmissionResult<T> {
+        return new BacktraceReportSubmissionResult<T>('Invalid token');
+    }
+
+    public static OnUnknownError<T>(message: string): BacktraceReportSubmissionResult<T> {
+        return new BacktraceReportSubmissionResult('Unknown', message);
+    }
+    public static OnNetworkingError<T>(message: string): BacktraceReportSubmissionResult<T> {
+        return new BacktraceReportSubmissionResult('Network Error', message);
```
adamcronin42, line 37: Out of curiosity, why do some of these pass the type `T` to the `BacktraceReportSubmissionResult` and others not?

adamcronin42 (approved):
Looks good!

packages/sdk-core/src/model/http/model/BacktraceSubmissionResult.ts
konraddysput: because of the generic function, the function definition should have a generic parameter, because the resulting object is generic. We can apply a default value to it and then the generic parameter will be redundant.

packages/node/src/BacktraceNodeRequestHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+    private getHttpClient(submissionUrl: URL) {
+        return submissionUrl.protocol === 'http' ? http : https;
+    }
+
+    private parseServerResponse<T>(response: string): T {
+        try {
+            return JSON.parse(response);
+        } catch {
+            return {} as T;
+        }
```
perf2711, line 131: Should we catch JSON failures? Maybe let the request catch handle this?

packages/browser/src/BacktraceBrowserRequestHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+                body: payload,
+                headers: typeof payload === 'string' ? this.JSON_HEADERS : {},
+                signal: controller.signal,
+            });
+
+            clearInterval(id);
+
+            switch (response.status) {
+                case 200: {
+                    const result = response.json() as T;
```
perf2711, line 54: You should `await` this.

packages/node/src/BacktraceNodeRequestHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+        const formData = new FormData();
+        formData.append(this.UPLOAD_FILE_NAME, json, `${this.UPLOAD_FILE_NAME}.json`);
+
+        if (!attachments || attachments.length === 0) {
+            return formData;
+        }
+
+        for (const attachment of attachments) {
+            if (typeof attachment === 'string') {
+                if (!fs.existsSync(attachment)) {
```
perf2711, line 143: Is there a reason why we're not using the promise/callback version of this?

perf2711 (changes requested):
Must do: `await` the json response in fetch.

packages/browser/src/BacktraceBrowserRequestHandler.ts
konraddysput: done, thank you! I wonder why eslint didn't complain about that.

packages/node/src/BacktraceNodeRequestHandler.ts
konraddysput: I still believe we're still figuring out the minimum supported version, this is why I preferred to use the API that definitely works everythwere.

packages/node/src/BacktraceNodeRequestHandler.ts
konraddysput: I had mixed feelings about that because the report is there and with callback apis, there is always a chance we won't be fast enough. This is why I wanted to still anyway tell user everything went ok, but the output might be unexpected. 


I changed the implementation and allow now parsing errors. We will revisit it once we recognize this is the issue for our users

packages/node/src/BacktraceNodeRequestHandler.ts
perf2711: Fair enough. But you could still promisify the callback version.

perf2711 (approved):

PR #8, opened 2023-06-12
Add helpers for future sourcemap plugins and CLI.

perf2711:
This PR adds `sourcemap-tools` package with `DebugIdGenerator` and `ContentAppender` classes to help create future plugins for working with sourcemaps.

Specific usage will be shown in the next PR with the Webpack plugin.

### DIFF
--- a/tools/sourcemap-tools/package.json
+++ b/tools/sourcemap-tools/package.json
@@ -1,0 +1,40 @@
+{
+    "name": "@backtrace/sourcemap-tools",
+    "version": "0.0.1",
+    "description": "Backtrace-JavaScript sourcemap tools",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "build": "tsc",
+        "clean": "tsc -b --clean && rimraf \"lib\"",
+        "format": "prettier --write '**/*.ts'",
+        "lint": "eslint . --ext .ts",
+        "watch": "tsc -w",
+        "test": "NODE_ENV=test jest"
+    },
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/backtrace-labs/backtrace-javascript.git"
+    },
+    "keywords": [
+        "Error",
+        "Reporting",
+        "Diagnostic",
+        "Tool",
+        "Bug",
+        "Bugs",
+        "StackTrace"
+    ],
+    "author": "Backtrace <team@backtrace.io>",
+    "license": "MIT",
+    "bugs": {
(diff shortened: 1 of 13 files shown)


tools/sourcemap-tools/package.json
```diff
@@ -0,0 +1,40 @@
+{
+    "name": "@backtrace/sourcemap-tools",
```
konraddysput, line 2: can we confirm the package name with Rick and Peter?

tools/sourcemap-tools/package.json
```diff
@@ -0,0 +1,40 @@
...
+        "url": "git+https://github.com/backtrace-labs/backtrace-javascript.git"
+    },
+    "keywords": [
+        "Error",
+        "Reporting",
+        "Diagnostic",
+        "Tool",
+        "Bug",
+        "Bugs",
+        "StackTrace"
```
konraddysput, line 26: can we add 

 "source maps",
    "sourcemaps"

tools/sourcemap-tools/package.json
```diff
@@ -0,0 +1,40 @@
+{
+    "name": "@backtrace/sourcemap-tools",
+    "version": "0.0.1",
```
konraddysput, line 3: Do we support all node envs? Otherwise we need engines support definition here

tools/sourcemap-tools/src/ContentAppender.ts
```diff
@@ -0,0 +1,11 @@
+export class ContentAppender {
+    public appendToJSON(json: string, keyValues: object) {
```
konraddysput, line 2: why we don't want to deserialize the json object and serialize it once again with new keys?

konraddysput (approved):
Leaving a few comments. The pull request looks really good to me. Lets just confirm everything and we can land it to dev

tools/sourcemap-tools/package.json
perf2711: Added.

tools/sourcemap-tools/src/ContentAppender.ts
konraddysput: wouldn't it be better if we do it once and keep the JSON file in memory for the source map generation?

melekr (approved):

tools/sourcemap-tools/src/ContentAppender.ts
perf2711: The source map is already generated at this point. I do not generate it - I only receive string concent.

konraddysput (approved):

adamcronin42 (approved):
Looks good!

PR #11, opened 2023-06-14
Rate limiter

konraddysput:
# Why

This diff adds rate limit watcher support to Backtrace Client. What does it mean? Sometimes some bugs generate reports in a loop. Because of that, the client app can flood Backtrace with data. Because of client limitations related to storage, the number of monthly reports, our clients need the option to prevent error loops.

This diff adds report limit watcher. The class doesn't care about report information but checks how many reports did we generate. In addition to that, the report might be with the old timestamp. in this situation, we should always trust the report limit watcher.


packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
```diff
@@ -0,0 +1,60 @@
+import { TimeHelper } from '../../common/TimeHelper';
+
+export class RateLimitWatcher {
+    /**
+     * Time the single report can stay in the queue.
+     */
+    public MAXIMUM_TIME_IN_QUEUE = 60;
```
perf2711, line 7: Can we use milliseconds here?

packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
```diff
@@ -0,0 +1,60 @@
...
+
+        // the queue can be cleaned up
+        if (time - this._reportQueue[this._reportQueue.length - 1] > this.MAXIMUM_TIME_IN_QUEUE) {
+            // if the last report is longer than it should
+            // we know we can clear the queue
+            this._reportQueue.length = 0;
+            return;
+        }
+
+        this._reportQueue = this._reportQueue.filter((n) => time - n < this.MAXIMUM_TIME_IN_QUEUE);
```
perf2711, line 58: The time array is always sorted ascending, right?

Maybe find the first time that is lower than passed time, and remove everything up to that time?

packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
perf2711, line 7: This is not a constant, is it? Why it is uppercase? Or maybe this should be `readonly` and `private`?

konraddysput: our reports use timestamps in sec required by Backtrace to send successful reports. I prefer to stay to this instead of adding ms here.

packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
konraddysput: This is what the condition above does :)

packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
konraddysput: it should be readonly. Thanks. The constant can be visible so we can do nice tests around this

packages/sdk-core/src/modules/rateLimiter/RateLimitWatcher.ts
perf2711: Nope, you're checking whether the whole queue can be cleaned up. Right here I'm talking about finding an index in the middle of the array that determines which times can be removed.

adamcronin42 (approved):

PR #14, opened 2023-06-16
Browser attributes

konraddysput:
# Why

This diff adds built-in web browser attributes to the backtrace-browser package.


packages/browser/src/attributes/UserIdentifierAttributeProvider.ts
```diff
@@ -0,0 +1,28 @@
...
+                window.localStorage.setItem(this.USER_IDENTIFIER, guid);
+            }
+            this._guid = guid;
+        }
+
+        return {
+            guid: this._guid,
+        };
+    }
+}
```
perf2711, line 28: Doesn't this fall under some consents about user privacy/tracking?

gittheking (approved):

packages/browser/src/attributes/WebsiteAttributeProvider.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const PAGE_START_TIME = TimeHelper.timeNowInMs();
+
+export class WebsiteAttributeProvider implements BacktraceAttributeProvider {
+    public get type(): 'scoped' | 'dynamic' {
+        return 'scoped';
+    }
+    public get(): Record<string, unknown> {
+        return {
+            application: document.title || 'unknown', // application is required. Using unknown string if it is not found.
```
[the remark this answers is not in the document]
konraddysput: Yes there is an option. It will be overriden by the client attributes

packages/browser/src/attributes/UserIdentifierAttributeProvider.ts
konraddysput: We work on this in the past when we added this to Backtrace-js. The user has an option to not use it

packages/browser/src/attributes/UserIdentifierAttributeProvider.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    public get type(): 'scoped' | 'dynamic' {
+        return 'scoped';
+    }
+    public get(): Record<string, unknown> {
+        if (!this._guid) {
+            let guid = window.localStorage.getItem(this.USER_IDENTIFIER);
+            if (!guid) {
+                guid = IdGenerator.uuid();
+                window.localStorage.setItem(this.USER_IDENTIFIER, guid);
```
adamcronin42, line 19: We may want to consider setting localStorage in a try catch as an exception can be thrown if the user configures the browser to not persist data. [https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#:~:text=Thrown%20in%20one,from%20persisting%20data.](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#:~:text=Thrown%20in%20one,from%20persisting%20data.)

konraddysput: Added thanks!

perf2711 (approved):

PR #18, opened 2023-06-19
Read package information

konraddysput:
# Why

This diff allows us to read the package.json file based on the require.main (might not exist in some different javascript modules) and process.cwd(). 

Application and version attributes are required. They must be defined to use a metrics system and report data correctly. This is why we also have loud information about missing package information in the attribute system.


packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+            application: this._applicationInformation['application'] ?? applicationData['name'],
+            ['application.version']: this._applicationInformation['application.version'] ?? applicationData['version'],
+        };
+    }
+
+    private readApplicationInformation(): Record<string, unknown> | undefined {
+        const maximumDepth = 5;
+        const possibleSourcePaths = [process.cwd()];
+        if (require.main?.path) {
+            possibleSourcePaths.unshift(path.resolve(require.main.path, '..'));
```
perf2711, line 43: `path.resolve(require.main.path, '..')` can be changed to `path.dirname(require.main.path)`.

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        for (let possibleSourcePath of possibleSourcePaths) {
+            for (let index = 0; index < maximumDepth; index++) {
+                const packagePath = path.join(possibleSourcePath, 'package.json');
+                if (!fs.existsSync(packagePath)) {
+                    possibleSourcePath = path.resolve(possibleSourcePath, '..');
+                    continue;
+                }
+                return JSON.parse(fs.readFileSync(packagePath, 'utf8'));
+            }
```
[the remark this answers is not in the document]
konraddysput: The reason why I added the maximumDepth is because I worried about the directory structure somewhere. I wanted to avoid a situation we will read seriously outside the package scope.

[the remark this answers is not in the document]
konraddysput: Removed maximum depth and I'm checking if the resolved directory is the same. Thank you!

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+            const packagePath = path.join(possibleSourcePath, 'package.json');
+            if (!fs.existsSync(packagePath)) {
+                const parentPath = path.dirname(possibleSourcePath);
+                // avoid checking the same directory twice.
+                if (parentPath === possibleSourcePath) {
+                    break;
+                }
+                possibleSourcePath = parentPath;
+                continue;
+            }
```
perf2711, line 63: I think you're missing an inner loop here descending lower and lower into parent directories.

konraddysput: yeah I was. Now we're leaving once we reach the root directory

packages/node/tests/attributes/applicationInformationAttributeProviderTests.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        it('Should try to find a package.json in the current project', () => {
+            const provider = new ApplicationInformationAttributeProvider({} as BacktraceConfiguration);
+            const attributes = provider.get();
+
+            expect(attributes[provider.APPLICATION_ATTRIBUTE]).toBe(expectedPackageJson.name);
+
+            expect(attributes[provider.APPLICATION_VERSION_ATTRIBUTE]).toBe(expectedPackageJson.version);
+        });
+
+        it('Should throw an error when the package.json information does exist', () => {
```
adamcronin42, line 51: *does not

adamcronin42 (approved):

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -3,65 +3,72 @@ import fs from 'fs';
...
-        for (let possibleSourcePath of possibleSourcePaths) {
-            const packagePath = path.join(possibleSourcePath, 'package.json');
-            if (!fs.existsSync(packagePath)) {
-                const parentPath = path.dirname(possibleSourcePath);
-                // avoid checking the same directory twice.
-                if (parentPath === possibleSourcePath) {
-                    break;
+    private readApplicationInformation(): Record<string, unknown> | undefined {
+        for (let possibleSourcePath of this.applicationSearchPaths) {
+            while (possibleSourcePath !== '.') {
```
perf2711, line 60: Why this is necessary?

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -3,65 +3,72 @@ import fs from 'fs';
...
         };
     }
 
-    private readApplicationInformation(): Record<string, unknown> | undefined {
+    private generateDefaultApplicationSearchPaths() {
         const possibleSourcePaths = [process.cwd()];
         if (require.main?.path) {
             possibleSourcePaths.unshift(path.dirname(require.main.path));
         }
+        return possibleSourcePaths;
```
perf2711, line 55: I think we should make sure that the paths are absolute. Maybe map over them and return `path.resolve`?

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
konraddysput: this is something we get after walking thru the whole directory structure when we cannot get the parent directory. Simply we come back to the current dir.

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
konraddysput: require.main.path is an absolute path: https://nodejs.org/api/modules.html#requiremain 

process.cwd() also returns the absolute path.

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
perf2711: Well, this is the same for `/` then and `c:/` or `d:/` on Windows?

I think either you should check those, or do this differently.

packages/node/tests/attributes/applicationInformationAttributeProviderTests.spec.ts
konraddysput: fixed. thx!

PR #24, opened 2023-06-22
Add verbosity to Webpack.

perf2711:
Adds logging to Webpack so we know what is going on.

Also adds test Webpack configuration with the plugin used to test the output.

### DIFF
--- a/tools/webpack-plugin/src/BacktracePluginV4.ts
+++ b/tools/webpack-plugin/src/BacktracePluginV4.ts
@@ -3,6 +3,8 @@ import crypto from 'crypto';
 import path from 'path';
 import { Compiler, WebpackPluginInstance } from 'webpack';
 import { BacktraceWebpackSourceGenerator } from './BacktraceWebpackSourceGenerator';
+import { statsPrinter } from './helpers/statsPrinter';
+import { AssetStats } from './models/AssetStats';
 import { BacktracePluginOptions } from './models/BacktracePluginOptions';
 
 export class BacktracePluginV4 implements WebpackPluginInstance {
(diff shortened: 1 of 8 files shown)


adamcronin42 (approved):

tools/webpack-plugin/src/BacktracePluginV4.ts
```diff
@@ -20,62 +22,113 @@ export class BacktracePluginV4 implements WebpackPluginInstance {
...
+                    logger.time(`[${key}] inject sourcemap key`);
+                    try {
+                        source = this._sourceGenerator.addDebugIdCommentToSource(
+                            source as never,
+                            debugId,
+                        ) as typeof source;
+                        logger.timeEnd(`[${key}] inject sourcemap key`);
+                        stats.sourceComment = true;
+                    } catch (err) {
+                        stats.sourceComment = err instanceof Error ? err : new Error('Unknown error.');
```
konraddysput, line 56: the time command from line 47 will never end. Should we add an error log about that? RIght now the user needs to check the source map comment however no one will do that in practice

tools/webpack-plugin/src/BacktracePluginV4.ts
```diff
@@ -20,62 +22,113 @@ export class BacktracePluginV4 implements WebpackPluginInstance {
...
+                    logger.time(`[${key}] append sourcemap key`);
+                    try {
+                        source = this._sourceGenerator.addDebugIdToRawSourceMap(
+                            source as never,
+                            stats.debugId,
+                        ) as never;
+                        logger.timeEnd(`[${key}] append sourcemap key`);
+                        stats.sourceMapAppend = true;
+                    } catch (err) {
+                        stats.sourceMapAppend = err instanceof Error ? err : new Error('Unknown error.');
```
konraddysput, line 75: the time command from line 66 will not end. Should we add a log error about that?

tools/webpack-plugin/src/helpers/statsPrinter.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (typeof stat === 'boolean') {
+        return stat ? 'successful' : 'skipped';
+    }
+
+    return stat.message;
+}
+
+export function statsPrinter(logger: webpack.Compilation['logger']) {
+    return function printStats(key: string, stats: AssetStats) {
+        const errors = [stats.sourceComment, stats.sourceMapAppend, stats.sourceMapUpload, stats.sourceSnippet].some(
```
konraddysput, line 18: how about strings? sourceSnippet can be string

tools/webpack-plugin/src/models/AssetStats.ts
```diff
@@ -0,0 +1,9 @@
+import { UploadResult } from '@backtrace/sourcemap-tools';
+
+export interface AssetStats {
+    readonly debugId: string;
+    sourceSnippet?: boolean | string | Error;
```
konraddysput, line 5: 1. why we should continue generating source maps information if we receive an error? 
2. why are we making this object so hard to maintain. Why do we want to print logs in the end instead of showing them when they happened? It's not natural. We can show the summary however a list of errors in the end might be huge if sourcemap cannot work for some kind of reason (let's say permission denied)

tools/webpack-plugin/src/models/AssetStats.ts
```diff
@@ -0,0 +1,9 @@
+import { UploadResult } from '@backtrace/sourcemap-tools';
+
+export interface AssetStats {
```
konraddysput, line 3: Let's think about what we have here. The purpose of the debugId property is to generate an output for the user in the source map. Why are we mixing it together with errors etc. It doesnt' sound right to me

konraddysput (commented):
I have mixed feelings about the design and decisions made in the logging system. For me it's not natural to spam error logs in the end instead of showing them when they happened. Why we want to still add a source map or continue adding source maps if we fail somewhere?

perf2711:
Closing in favor of https://github.com/backtrace-labs/backtrace-javascript/pull/29

PR #25, opened 2023-06-22
Adding react package to the monorepo

adamcronin42:
Adding the shell for the React package for our SDK

- Creating the package
- Updating package.json
- Setting up typescript
- Updating tsconfigs
- Copying webpack functionality
- License


packages/react/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from '@backtrace/browser';
```
konraddysput, line 1: I have mixed feelings about that - I know we want to have something that compiles and generates an output, however the first thing that we need to do just after adding anything is to remove this line of code

konraddysput (approved):

perf2711 (approved):

PR #27, opened 2023-06-23
Node.js example

konraddysput:
# Why

This diff adds a nodejs example to the repository. In addition to that we have also debug task.


examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    output: process.stdout,
+});
+
+const client = BacktraceClient.builder({
+    url: SUBMISSION_URL,
+    attachments: [path.join(path.dirname(process.cwd()), 'samplefile.txt')],
+    rateLimit: 5,
+    userAttributes: {
+        'custom-attribute': 'test',
+        'complex-attribute': {
```
perf2711, line 18: This will be used as an annotation, right? Maybe we should not confuse the user and name this `custom-annotation`?

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+            prop1: true,
+            prop2: 123,
+        },
+    },
+}).build();
+
+console.log(
+    `Welcome in the Backtrace demo. Please pick one of available options
```
perf2711, line 26: Should be "Welcome to the Backtrace demo".

perf2711 (commented):
Do we need an elaborate menu to do this example?

Why not just create two files that, when executed, run these examples:
* `sendReport.ts`
* `sendMessage.ts`

konraddysput:
> Do we need an elaborate menu to do this example?

Yes, because we will have more examples of our SDK than sending reports/messages. We will have soon unhandled exceptions/metrics/breadcrumbs/crashes. Most our SDKs in the demo have an option to choose to see the example behavior

perf2711:
Yeah, but most of the SDKs don't have the liberty of having multiple entry points, hence the menu requirement. Here we can run any file that we want as an entry point.

examples/sdk/node/src/index.ts
konraddysput: changed. Thx

examples/sdk/node/src/index.ts
konraddysput: Changed. Thx

konraddysput:
> Yeah, but most of the SDKs don't have the liberty of having multiple entry points, hence the menu requirement. Here we can run any file that we want as an entry point.

yeah but why it matters? The logic is simple and will be simple. Dividing it into multiple files will hide more than it will give us. As a developer of a super silly and straightforward logic, I would rather to see everything in one file if the logic in the file is just one single line of code. It doesn't make sense to me.

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+        const attributes = { selectedOption: option };
+
+        switch (option) {
+            case 1: {
+                try {
+                    fs.readFileSync('/path/to/not/existing/file');
+                } catch (err) {
+                    await client.send(err as Error, attributes);
+                }
```
perf2711, line 48: Can we extract all of cases to separate functions?

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    0. Exit`,
+);
+
+function showMenu() {
+    reader.question('Select the option... \n', async function executeUserOption(optionString: string) {
+        const option = parseInt(optionString);
+        if (isNaN(option)) {
+            console.error('Selected invalid option');
+            return exit(1);
```
perf2711, line 38: Why do we exit here?

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+console.log(
+    `Welcome to the Backtrace demo. Please pick one of available options
+    1. Send an exception
+    2. Send a message
+
+    0. Exit`,
+);
+
+function showMenu() {
+    reader.question('Select the option... \n', async function executeUserOption(optionString: string) {
```
perf2711, line 34: Can we move the console log from lines 25-30 to this question? Make sure to also specify that you want the user to input a number.

Also, IMO it would be nicer if we would not add a newline and instead do something like this:
```
Select the option: <user input here>
```

perf2711 (commented):
If we're going for a menu, maybe let's do it nice and use this for example: https://github.com/SBoudrias/Inquirer.js ?

examples/sdk/node/src/index.ts
konraddysput: I think it would be better if we handle NaN via default flow in the menu. Removed it. Thanks.

examples/sdk/node/src/index.ts
konraddysput: Moved it to the question. I also left the greeting message at the top of the menu, it sounds silly to show it every time.

examples/sdk/node/src/index.ts
konraddysput: I moved all code to methods.

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+}).build();
+
+console.log('Welcome to the @Backtrace/node demo');
+const menu = `Please pick one of available options:
+1. Send an exception
+2. Send a message
+0. Exit
+    
+Type the option number:`;
```
perf2711, line 31: Can we move this variable to `showMenu` function?

examples/sdk/node/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        switch (option) {
+            case 1: {
+                await sendHandledException(attributes);
+                break;
+            }
+            case 2: {
+                await sendMessage('test message', attributes);
+                break;
+            }
```
perf2711, line 61: You can do `return await sendHandledException(attributes);` and skip `break;` and braces. Same in `case 2:`.

perf2711 (approved):

examples/sdk/node/src/index.ts
konraddysput: done.

examples/sdk/node/src/index.ts
konraddysput: done.

PR #28, opened 2023-06-26
Add `SourceProcessor` class

perf2711:
When testing Webpack output with potential SDK changes, I've stumbled upon an error - source snippets adding debug IDs may have been executed later than the report was created, thus breaking the whole debug ID idea.

So I've opted to add them to the beginning of the file. This, however, breaks the sourcemaps, because a new line is added and all mappings are now offset.

I have decided to create a bundler-agnostic solution that appends the necessary code, and then fixes the sourcemaps. Webpack changes will be added in the next PR.

### DIFF
--- a/tools/sourcemap-tools/src/SourceProcessor.ts
+++ b/tools/sourcemap-tools/src/SourceProcessor.ts
@@ -6,6 +6,13 @@ import { stringToUuid } from './helpers/stringToUuid';
 export class SourceProcessor {
     constructor(private readonly _debugIdGenerator: DebugIdGenerator) {}
 
+    /**
+     * Adds required snippets and comments to source, and modifies sourcemap to include debug ID.
+     * @param source Source content.
+     * @param sourceMap Sourcemap object or JSON.
+     * @param debugId Debug ID. If not provided, one will be generated from `source`.
+     * @returns Used debug ID, new source and new sourcemap.
+     */
     public async processSourceAndSourceMap(source: string, sourceMap: string | RawSourceMap, debugId?: string) {
         if (!debugId) {
             debugId = stringToUuid(source);
@@ -23,6 +30,14 @@ export class SourceProcessor {
         return { debugId, source: newSource, sourceMap: newSourceMap };
     }
 
+    /**
+     * Adds required snippets and comments to source, and modifies sourcemap to include debug ID.
+     * Will write modified content to the files.
+     * @param sourcePath Path to the source.
+     * @param sourceMapPath Path to the sourcemap.
+     * @param debugId Debug ID. If not provided, one will be generated from `source`.
+     * @returns Used debug ID.
+     */
     public async processSourceAndSourceMapFiles(sourcePath: string, sourceMapPath: string, debugId?: string) {
         const source = await fs.promises.readFile(sourcePath, 'utf8');
         const sourceMap = await fs.promises.readFile(sourceMapPath, 'utf8');
--- a/tools/sourcemap-tools/src/SourceProcessor.ts
+++ b/tools/sourcemap-tools/src/SourceProcessor.ts
@@ -24,7 +24,13 @@ export class SourceProcessor {
         const newSource = sourceSnippet + '\n' + source + '\n' + sourceComment;
 
         // We need to offset the source map by amount of lines that we're inserting to the source code
-        const offsetSourceMap = await this.offsetSourceMap(sourceMap, 0, 1);
+        // Sourcemaps map code like this:
+        // original code X:Y => generated code A:B
+        // So if we add any code to generated code, mappings after that code will become invalid
+        // We need to offset the mapping lines by sourceSnippetNewlineCount:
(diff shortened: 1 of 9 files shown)


tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        consumer.eachMapping((m) => {
+            if (m.generatedLine < fromLine) {
+                return;
+            }
+
+            newSourceMap.addMapping({
+                source: m.source,
+                name: m.name,
+                generated:
```
konraddysput, line 57: why the interface don't accept undefined for genearted/original? Instead of using type hacks you can just pass undefined. In your definition you said "It's a position!" however it isn't a position

tools/sourcemap-tools/src/helpers/bytesToUuid.ts
```diff
@@ -0,0 +1,13 @@
+export function bytesToUuid(bytes: Buffer) {
+    return (
```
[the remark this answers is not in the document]
perf2711: This is not UUIDv4. This function only converts the buffer to a string UUID.

If you're referring to `stringToUuid` function, it generates a UUID based on SHA-1 hash. So it's not random as well.

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: Funny thing is that this interface does not specify `null` as an available option, but requires you to pass it when `generatedColumn` or `generatedLine` are `null`. So I cast it here. I'll add a comment about this.

tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -0,0 +1,71 @@
+import fs from 'fs';
+import { BasicSourceMapConsumer, Position, RawSourceMap, SourceMapConsumer, SourceMapGenerator } from 'source-map';
+import { DebugIdGenerator } from './DebugIdGenerator';
+import { stringToUuid } from './helpers/stringToUuid';
+
+export class SourceProcessor {
+    constructor(private readonly _debugIdGenerator: DebugIdGenerator) {}
+
+    public async processSourceAndSourceMap(source: string, sourceMap: string | RawSourceMap, debugId?: string) {
```
konraddysput, line 9: what is source? Is this the path to source map? Is this the source map content? I checked the code however it makes sense to make leave the note

tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (!debugId) {
+            debugId = stringToUuid(source);
```
konraddysput, line 11: Let's pretend I have a colossal javascript file—1000 lines of code. I'm changing the last one.  If I uploaded source maps before the change, will this generate the same uuid after my change?

tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        const sourceSnippet = this._debugIdGenerator.generateSourceSnippet(debugId);
+        const sourceComment = this._debugIdGenerator.generateSourceComment(debugId);
+
+        const newSource = sourceSnippet + '\n' + source + '\n' + sourceComment;
+
+        // We need to offset the source map by amount of lines that we're inserting to the source code
```
konraddysput, line 19: because...

konraddysput (commented):
Before we land - can you please check my question about uuid? we can't generate the same uuid twice.

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: `source` is the source file content. `sourceMap` is the source map JSON or object.

Will add some docs.

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: Sourcemaps map original position to generated position like this:
`original line X, original column Y => generated line A, generated column B`

If I add anything to the generated source, the sourcemap will be invalid, because generated lines will be changed. So if I add a line at the beginning of the file, I need to offset them like this:
`original line X, original column Y => generated line A + 1, generated column B`

tools/sourcemap-tools/src/SourceProcessor.ts
konraddysput: ah ok got it. I missed that part.

tools/sourcemap-tools/src/SourceProcessor.ts
konraddysput: ah ok got it - so +1 is because of our code injected to each map. Got it. Does it make sense to add it to the comment?

perf2711: I expanded the comment.

adamcronin42 (approved):

PR #29, opened 2023-06-26
Use `SourceProcessor` in Webpack

perf2711:
This PR changes how Webpack works - it will now only run the `SourceProcessor` class to change the emitted files.

Due to this change, it is no longer necessary to maintain two versions of Webpack plugins, so I have merged them into one.


tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -1,19 +1,77 @@
...
+import { BacktracePluginOptions } from './models/BacktracePluginOptions';
+
+export class BacktracePlugin implements WebpackPluginInstance {
+    private readonly _sourceMapProcessor: SourceProcessor;
+    private readonly _sourceMapUploader?: SourceMapUploader;
+
+    constructor(public readonly options?: BacktracePluginOptions) {
+        this._sourceMapProcessor = new SourceProcessor(options?.debugIdGenerator ?? new DebugIdGenerator());
+
+        this._sourceMapUploader =
```
konraddysput, line 13: I didn't get answer to my question about this in the previous diff. How the source map uploader can be undefined? What is the purpose of having our plugin without source map uploader?

perf2711: Well, you could always upload the files manually, if you want. The debug IDs will still be injected.

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -1,19 +1,77 @@
...
+                if (this._sourceMapUploader) {
+                    logger.time(`[${asset}] upload sourcemap`);
+                    try {
+                        await this._sourceMapUploader.upload(sourceMapPath, debugId);
+                        logger.timeEnd(`[${asset}] upload sourcemap`);
+                    } catch (err) {
+                        logger.timeEnd(`[${asset}] upload sourcemap`);
+                        logger.error(`[${asset}] upload sourcemap failed:`, err);
+                    }
+                    logger.log(`[${asset}] file processed and sourcemap uploaded`);
```
adamcronin42, line 70: Would we want to change the phrasing as sourcemap upload may have failed?

adamcronin42 (approved):
Looks good! Left one small comment

tools/webpack-plugin/src/models/BacktracePluginOptions.ts
```diff
@@ -1,5 +1,7 @@
-import { DebugIdGenerator } from '@backtrace/sourcemap-tools';
+import { DebugIdGenerator, SourceMapUploader } from '@backtrace/sourcemap-tools';
 
 export interface BacktracePluginOptions {
     debugIdGenerator?: DebugIdGenerator;
+    sourceMapUploader?: SourceMapUploader;
```
konraddysput, line 5: This is the customer-facing interface - can we describe all options and what it means if something is not set up - for example, "uploadUrl"? As a user, I'm not sure how the library will behave when uploadUrl is undefined.

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -1,19 +1,77 @@
...
+
+        this._sourceMapUploader =
+            options?.sourceMapUploader ?? (options?.uploadUrl ? new SourceMapUploader(options.uploadUrl) : undefined);
+    }
+
+    public apply(compiler: webpack.Compiler) {
+        compiler.hooks.afterEmit.tapPromise(BacktracePlugin.name, async (compilation) => {
+            const logger = compilation.getLogger(BacktracePlugin.name);
+            if (!compilation.outputOptions.path) {
+                logger.error('skipping everything because outputOptions.path is not set, a bug?');
```
konraddysput, line 21: 1. Skipping
2. "The outputOptions.path is not defined. Cannot generate source maps for not existing path" Isn't this description better?

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -1,19 +1,77 @@
...
+                try {
+                    debugId = await this._sourceMapProcessor.processSourceAndSourceMapFiles(sourcePath, sourceMapPath);
+                    logger.timeEnd(`[${asset}] process source and sourcemap`);
+                } catch (err) {
+                    logger.timeEnd(`[${asset}] process source and sourcemap`);
+                    logger.error(`[${asset}] process source and sourcemap failed:`, err);
+                    continue;
+                }
+
+                if (this._sourceMapUploader) {
```
konraddysput, line 61: you can save one more tab if you reverse the if statement and use continue

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -1,19 +1,77 @@
...
+                    logger.time(`[${asset}] upload sourcemap`);
+                    try {
+                        await this._sourceMapUploader.upload(sourceMapPath, debugId);
+                        logger.timeEnd(`[${asset}] upload sourcemap`);
```
konraddysput, line 65: wouldn't it be better if you put timeEnd in finally block?

konraddysput (changes requested):
Can we please add description to the customer-facing interface options and describe behavior when something is not set?

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: Generally, Webpack does not allow to configure it without an output path. So this theoretically should never happen. Even the type docs describe this as required, though for some reason, it is marked as optional.

image

tools/webpack-plugin/src/models/BacktracePluginOptions.ts
perf2711: I added some docs.

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: Done

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: I guess this could work. Done

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: Good catch, done.

konraddysput (approved):

konraddysput (approved):

PR #32, opened 2023-06-27
Add printing asset stats to Webpack plugin

perf2711:
This PR adds a summary of modified and uploaded files at the end of the compilation.

### DIFF
--- a/tools/webpack-plugin/src/BacktracePlugin.ts
+++ b/tools/webpack-plugin/src/BacktracePlugin.ts
@@ -1,6 +1,8 @@
 import { DebugIdGenerator, SourceMapUploader, SourceProcessor } from '@backtrace/sourcemap-tools';
 import path from 'path';
 import webpack, { WebpackPluginInstance } from 'webpack';
+import { statsPrinter } from './helpers/statsPrinter';
+import { AssetStats } from './models/AssetStats';
 import { BacktracePluginOptions } from './models/BacktracePluginOptions';
 
 export class BacktracePlugin implements WebpackPluginInstance {
@@ -15,6 +17,8 @@ export class BacktracePlugin implements WebpackPluginInstance {
     }
 
     public apply(compiler: webpack.Compiler) {
+        const assetStats = new Map<string, AssetStats>();
+
         compiler.hooks.afterEmit.tapPromise(BacktracePlugin.name, async (compilation) => {
             const logger = compilation.getLogger(BacktracePlugin.name);
             if (!compilation.outputOptions.path) {
@@ -48,13 +52,20 @@ export class BacktracePlugin implements WebpackPluginInstance {
             logger.log(`received ${entries.length} files for processing`);
 
             for (const [asset, sourcePath, sourceMapPath] of entries) {
+                const stats: AssetStats = {};
+                assetStats.set(asset, stats);
+
                 let debugId: string;
 
                 logger.time(`[${asset}] process source and sourcemap`);
                 try {
                     debugId = await this._sourceProcessor.processSourceAndSourceMapFiles(sourcePath, sourceMapPath);
+                    stats.debugId = debugId;
+                    stats.processSource = true;
+                    logger.timeEnd(`[${asset}] process source and sourcemap`);
                 } catch (err) {
                     logger.error(`[${asset}] process source and sourcemap failed:`, err);
+                    stats.processSource = err instanceof Error ? err : new Error('Unknown error.');
                     continue;
                 } finally {
                     logger.timeEnd(`[${asset}] process source and sourcemap`);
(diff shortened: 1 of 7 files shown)


tools/webpack-plugin/src/models/AssetStats.ts
```diff
@@ -0,0 +1,7 @@
+import { UploadResult } from '@backtrace/sourcemap-tools';
+
+export interface AssetStats {
+    debugId?: string;
```
konraddysput, line 4: From a prototyping perspective, it's ok to have everything nullable, but is that the case in the end? The asset stats must have debugId. I would even say more - if we print anything to the user, the model also should have a path to a file for unexpected errors with unknown error messages from the user's point of view without a path to a field.

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -48,30 +52,40 @@ export class BacktracePlugin implements WebpackPluginInstance {
...
                     logger.error(`[${asset}] process source and sourcemap failed:`, err);
+                    stats.processSource = err instanceof Error ? err : new Error('Unknown error.');
                     continue;
                 }
 
                 if (this._sourceMapUploader) {
                     logger.time(`[${asset}] upload sourcemap`);
                     try {
-                        await this._sourceMapUploader.upload(sourceMapPath, debugId);
+                        const result = await this._sourceMapUploader.upload(sourceMapPath, debugId);
```
konraddysput, line 74: do we upload source map one by one instead of in an archive?

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -48,30 +52,40 @@ export class BacktracePlugin implements WebpackPluginInstance {
...
+                    stats.processSource = true;
                     logger.timeEnd(`[${asset}] process source and sourcemap`);
                 } catch (err) {
                     logger.timeEnd(`[${asset}] process source and sourcemap`);
                     logger.error(`[${asset}] process source and sourcemap failed:`, err);
+                    stats.processSource = err instanceof Error ? err : new Error('Unknown error.');
                     continue;
                 }
 
                 if (this._sourceMapUploader) {
```
konraddysput, line 71: is there any reason to upload source map without debugId?

tools/webpack-plugin/src/helpers/statsPrinter.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if (typeof stat === 'boolean') {
+        return stat ? 'successful' : 'skipped';
+    }
+
+    return stat.message;
+}
+
+export function statsPrinter(logger: webpack.Compilation['logger']) {
```
konraddysput, line 16: Let's think about similar use cases. When I think about what we need to do I think its fair to compare it with an output from for example jest. Why? During jest tests all the time you see incremental output about passed/failed tests. 

The user might stop them any time. The user in the end sees how many tests passed/failed. In the end when we want to see the summary.

Why we don't collect information about successes/failures and print information in real time. Uploading symbols consumes space. Maybe it's not the case right now, however, it will be a case in the future. The user wants to stop the uploading process asap when the user sees an issue.  

tl;dr;
* compare the output with jest - we spam too much redundant information in the end of the cli - for projects like workflows and unexpected error (let's say type error in our code somewhere) you will generate 1300 error messages in the end as a user I expect to see stats. 
* we can show errors/useful information in real-time to the user during processing and in the end just show stats.  The user already has everything needed to understand where is an error.
* I love to code, but we can avoid doing that just for simply incrementing success/failure variables

konraddysput (commented):
I really don't like the flow and what we show to the user. Can we please verify it together? I'm not requesting changes because I might miss something.

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: Right now, one by one. This is out of scope of this PR.

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: We do not upload a sourcemap without debugId - see `continue` on line 68.

tools/webpack-plugin/src/helpers/statsPrinter.ts
perf2711: These are Webpack guidelines. Webpack does not spam anything during the build process, and only prints stats at the end. If you enable more verbose logging, you will have compilation-time logs.

See https://webpack.js.org/api/logging/.

image

>you will generate 1300 error messages in the end as a user I expect to see stats.
If you output 1300 files through Webpack, something is definitely wrong with your setup. Still, Webpack's output will be 30x larger in this case than what we output.

The most I could do is not to print each file separately in default log mode, and display just a status of what has been done for all files.

tools/webpack-plugin/src/BacktracePlugin.ts
konraddysput: Yeah sure, this is definitely something that doesn't belong to this pull request, however, I saw the code and I asked a question ^^. Thank you!

adamcronin42 (approved):

konraddysput (approved):

melekr (approved):

PR #33, opened 2023-06-27
Add debug IDs support to sdk-core, node and browser packages

perf2711:
This PR adds support for debug IDs in previously built files.

How it works:
1. debug IDs are being injected while building the files using for example a Webpack plugin
2. when creating a report, debug IDs are loaded using `DebugIdProvider`:
    1. debug IDs are being retrieved using `DebugIdMapProvider` - in node and browser using `VariableDebugIdMapProvider` from `global` and `window`, respectively,
    2. first frame's library from each file's stack and debug ID are saved
3. the debug IDs are assigned to each frame to `debug_identifier` property where a debug ID for library exists

### DIFF
--- a/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
+++ b/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
@@ -24,7 +24,6 @@ export class BacktraceDataBuilder {
 
         const stackTrace = this._stackTraceConverter.convert(report.stackTrace, report.message);
 
-        this._debugIdProvider.loadDebugIds();
         for (const frame of stackTrace) {
             frame.debug_identifier = this._debugIdProvider.getDebugId(frame.library);
         }
--- a/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
+++ b/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
@@ -26,7 +26,7 @@ export class BacktraceDataBuilder {
 
         this._debugIdProvider.loadDebugIds();
         for (const frame of stackTrace) {
-            frame.debugId = this._debugIdProvider.getDebugId(frame.library);
+            frame.debug_identifier = this._debugIdProvider.getDebugId(frame.library);
         }
 
         return {
--- a/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
+++ b/packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
@@ -1,4 +1,4 @@
-import { BacktraceStackTraceConverter } from '../..';
+import { BacktraceStackTraceConverter, DebugIdProvider } from '../..';
 import { SdkOptions } from '../../builder/SdkOptions';
 import { IdGenerator } from '../../common/IdGenerator';
 import { TimeHelper } from '../../common/TimeHelper';
@@ -12,6 +12,7 @@ export class BacktraceDataBuilder {
     constructor(
         private readonly _sdkOptions: SdkOptions,
         private readonly _stackTraceConverter: BacktraceStackTraceConverter,
+        private readonly _debugIdProvider: DebugIdProvider,
     ) {}
 
     public build(
@@ -20,6 +21,13 @@ export class BacktraceDataBuilder {
         clientAnnotations: Record<string, unknown> = {},
     ): BacktraceData {
         const reportData = ReportDataBuilder.build(report.attributes);
+
+        const stackTrace = this._stackTraceConverter.convert(report.stackTrace, report.message);
+
+        this._debugIdProvider.loadDebugIds();
+        for (const frame of stackTrace) {
+            frame.debugId = this._debugIdProvider.getDebugId(frame.library);
+        }
 
         return {
             uuid: IdGenerator.uuid(),
@@ -34,7 +42,7 @@ export class BacktraceDataBuilder {
                 [this.MAIN_THREAD_NAME]: {
                     fault: true,
                     name: this.MAIN_THREAD_NAME,
-                    stack: this._stackTraceConverter.convert(report.stackTrace, report.message),
+                    stack: stackTrace,
                 },
             },
             annotations: {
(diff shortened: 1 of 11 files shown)


packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
```diff
@@ -21,6 +22,13 @@ export class BacktraceDataBuilder {
...
         const reportData = ReportDataBuilder.build(report.attributes);
 
+        const stackTrace = this._stackTraceConverter.convert(report.stackTrace, report.message);
+
+        this._debugIdProvider.loadDebugIds();
+        for (const frame of stackTrace) {
+            frame.debug_identifier = this._debugIdProvider.getDebugId(frame.library);
+        }
+
         return {
```
konraddysput, line 32: important note before we land: 

debug_identifier with our model always goes with the modules declaration in global annotations. The modules declaration allows us to match specific debug_id to the library stored in our symbols system. I know, it might work differently for source maps, however, for every other integration one piece of information requires the other.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+        const message = new Error().message;
+        const result: Record<string, string> = {};
+        for (const [rawStack, debugId] of Object.entries(debugIdMap)) {
+            const frames = this._stackTraceConverter.convert(rawStack, message);
+            if (!frames.length) {
+                continue;
+            }
+
+            // The first frame will have the file's path
+            const frame = frames[0];
```
konraddysput, line 35: how about unknown library where the library info is not available

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const SOURCE_DEBUG_ID_VARIABLE = '_btDebugIds';
+
+export class DebugIdProvider {
+    private _fileDebugIds?: Record<string, string>;
+
+    constructor(
+        private readonly _stackTraceConverter: BacktraceStackTraceConverter,
+        private readonly _debugIdMapProvider?: DebugIdMapProvider,
+    ) {}
```
konraddysput, line 12: why we cannot file fileDebugIds in the constructor? Initialization time is a great moment to do that, instead of doing it on demand when the record happened. 


Why?
We have limited time/resources when an exception occurs. In worst case scenario it might take a while to load it and later reprocess.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+import { DebugIdMapProvider } from './interfaces/DebugIdMapProvider';
+
+export const SOURCE_DEBUG_ID_VARIABLE = '_btDebugIds';
+
+export class DebugIdProvider {
+    private _fileDebugIds?: Record<string, string>;
+
+    constructor(
+        private readonly _stackTraceConverter: BacktraceStackTraceConverter,
+        private readonly _debugIdMapProvider?: DebugIdMapProvider,
```
konraddysput, line 11: wouldn't it be better if sdk will pass debug id map instead of provider? The provider should always be defined, otherwise, there is no reason of generating debug id map. I see how you mock your test cases, but I believe in the real case scenario outside tests we won't pass debugIdMap to loadDebguids (what is the reason to pass it if we already have them?)

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+            // The first frame will have the file's path
+            const frame = frames[0];
+            result[frame.library] = debugId;
+        }
+
+        return (this._fileDebugIds = result);
+    }
+
+    public getDebugId(file: string): string | undefined {
+        return this._fileDebugIds?.[file];
```
konraddysput, line 43: what if someone didn't load it first? we can easily make it less error prone if we move it to the constructor

konraddysput (changes requested):
Can we cache/load debug id map on the SDK startup? Loading it on demand might cause a lot of problems when the first error we're capturing is unhandled exception.


 Loading them on demand for unhandled exception cost time when the application stuck and doesn't do anything or can cause a data loss if node/system decides to kill us.

packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
perf2711: I talked with George and he required that frames only have debug_identifier fields.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
perf2711: This will never happen by design, the top-most frame will always have a file.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
perf2711: Changed.

packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
```diff
@@ -21,6 +22,13 @@ export class BacktraceDataBuilder {
     ): BacktraceData {
         const reportData = ReportDataBuilder.build(report.attributes);
 
+        const stackTrace = this._stackTraceConverter.convert(report.stackTrace, report.message);
+
+        this._debugIdProvider.loadDebugIds();
```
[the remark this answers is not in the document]
perf2711: Changed.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
perf2711: Changed.

packages/sdk-core/src/sourcemaps/DebugIdProvider.ts
perf2711: I want to make this as transparent as possible. If for some reason we do not support debug IDs, it should not fail because of missing debug ID map provider.

konraddysput (approved):

adamcronin42 (approved):

PR #34, opened 2023-06-27
Submission URL parser

konraddysput:
# Why

In our submission URLs we have a lot of important information. We should continue using them but also reuse information from them in other places of the SDK - for example, to learn what is the user universe name or the submission token. 

This diff adds URL parsers to javascript SDK

### DIFF
--- a/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
+++ b/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
@@ -1,0 +1,71 @@
+export class SubmissionUrlInformation {
+    private static SUBMIT_PREFIX = 'submit.backtrace.io/';
+
+    /**
+     * Convert url/token from credentials to JSON submission URL
+     * @param url credentials URL
+     * @param token credentials token
+     * @returns JSON submissionURL
+     */
+    public static toJsonReportSubmissionUrl(url: string, token?: string): string {
+        // if the token doesn't exist - use URL
+        if (!token) {
+            return url;
+        }
+
+        // if the url points to submit, we should always use it without any modifications
+        if (url.includes(this.SUBMIT_PREFIX)) {
+            return url;
+        }
+
+        // if the URL has token in the URL, the user probably added a token once again
+        // in this case, don't do anything
+        if (url.indexOf(token) !== -1) {
+            return url;
+        }
+
+        return new URL(`/post?format=json&token=${token}`, url).href;
+    }
+
+    /**
+     * Find the universe based on the submission URL
+     * @param submissionUrl submission URL
+     * @returns universe name
+     */
+    public static findUniverse(submissionUrl: string): string {
+        const submitIndex = submissionUrl.indexOf(this.SUBMIT_PREFIX);
+        if (submitIndex !== -1) {
+            const universeStartIndex = submitIndex + this.SUBMIT_PREFIX.length;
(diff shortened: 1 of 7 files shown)


packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        const tokenQueryParameter = 'token=';
+        const tokenQueryParameterIndex = submissionUrl.indexOf(tokenQueryParameter);
+        if (tokenQueryParameterIndex === -1) {
+            return undefined;
+        }
+
+        const tokenStartIndex = tokenQueryParameterIndex + tokenQueryParameter.length;
+        return submissionUrl.substring(tokenStartIndex, tokenStartIndex + tokenLength);
```
perf2711, line 69: You can create a new URL and retrieve the token from `searchParams`.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -0,0 +1,71 @@
...
+     * @param submissionUrl submission URL
+     * @returns universe name
+     */
+    public static findUniverse(submissionUrl: string): string {
+        const submitIndex = submissionUrl.indexOf(this.SUBMIT_PREFIX);
+        if (submitIndex !== -1) {
+            const universeStartIndex = submitIndex + this.SUBMIT_PREFIX.length;
+            const endOfUniverseName = submissionUrl.indexOf('/', universeStartIndex);
+            return submissionUrl.substring(universeStartIndex, endOfUniverseName);
+        }
```
perf2711, line 41: Can you add comments like below on how do you expect these URLs to look like?

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -0,0 +1,71 @@
...
+        // the universe name should be available in the hostname
+        // for example abc.sp.backtrace.io or zyx.in.backtrace.io or foo.backtrace.io
+        const hostname = new URL(submissionUrl).host;
+        const endOfUniverseName = hostname.indexOf('.');
+
+        return hostname.substring(0, endOfUniverseName);
```
perf2711, line 47: We should use this method only if the origin ends with `backtrace.io`. With other URLs, we cannot be sure that the first element is the universe name.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -0,0 +1,71 @@
...
+            return url;
+        }
+
+        // if the URL has token in the URL, the user probably added a token once again
+        // in this case, don't do anything
+        if (url.indexOf(token) !== -1) {
+            return url;
+        }
+
+        return new URL(`/post?format=json&token=${token}`, url).href;
```
perf2711, line 27: In that case, I would create the URL, and in `searchParams` set the `"token"` query param to `token`.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
konraddysput: After thinking once again about my assumption - trusting the hostname always to contain the universe at the beginning of the URL, I agree with you. We shouldn't trust domains other than backtrace.io here. Added a change

perf2711 (approved):

PR #35, opened 2023-06-27
Command line attribute provider

konraddysput:
# Why

This diff adds an additional way to determinate the application package.json path. This time, this diff adds support to node.js  files executed via command lines like node ./file.js

### DIFF
--- a/packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
+++ b/packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
@@ -54,7 +54,7 @@ export class ApplicationInformationAttributeProvider implements BacktraceAttribu
         if (potentialCommandLineStartupFile) {
             const potentialCommandLineStartupFilePath = path.resolve(potentialCommandLineStartupFile);
             if (fs.existsSync(potentialCommandLineStartupFilePath)) {
-                possibleSourcePaths.unshift(path.resolve(potentialCommandLineStartupFilePath));
+                possibleSourcePaths.unshift(potentialCommandLineStartupFilePath);
             }
         }
         if (require.main?.path) {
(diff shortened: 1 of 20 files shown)


packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -50,6 +50,10 @@ export class ApplicationInformationAttributeProvider implements BacktraceAttribu
 
     private generateDefaultApplicationSearchPaths() {
         const possibleSourcePaths = [process.cwd()];
+        const potentialCommandLineStartupFile = process.argv[1];
+        if (potentialCommandLineStartupFile && fs.existsSync(potentialCommandLineStartupFile)) {
+            possibleSourcePaths.unshift(potentialCommandLineStartupFile);
```
perf2711, line 55: Can we `path.resolve` this path before returning it?

adamcronin42: For my learning, why do we need an absolute path?

adamcronin42 (approved):

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
konraddysput: the relative path generated by the backtrace-node library might point to a different place if sdk-core starts using it.

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -51,8 +51,11 @@ export class ApplicationInformationAttributeProvider implements BacktraceAttribu
     private generateDefaultApplicationSearchPaths() {
         const possibleSourcePaths = [process.cwd()];
         const potentialCommandLineStartupFile = process.argv[1];
-        if (potentialCommandLineStartupFile && fs.existsSync(potentialCommandLineStartupFile)) {
-            possibleSourcePaths.unshift(potentialCommandLineStartupFile);
+        if (potentialCommandLineStartupFile) {
+            const potentialCommandLineStartupFilePath = path.resolve(potentialCommandLineStartupFile);
+            if (fs.existsSync(potentialCommandLineStartupFilePath)) {
+                possibleSourcePaths.unshift(path.resolve(potentialCommandLineStartupFilePath));
+            }
```
perf2711, line 58: Why are you resolving this twice?

perf2711 (approved):

PR #36, opened 2023-06-27
Feature/metrics

konraddysput:
# Why

This diff adds metrics support to all our clients. The metrics system works differently depending on the environment we use it. In the web browser, the same application can be opened in multiple web browser tabs. In node, we have a single session we can manage.

This diff was created based on the code and decisions made in the past for backtrace-js/backrace-node metrics implementation. Decisions made by the team to store guid/last active date are reflected in this code.


packages/sdk-core/src/modules/metrics/MetricsBuilder.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (!uniqueEventsSubmissionQueue) {
+            return undefined;
+        }
+
+        summedEventsSubmissionQueue =
+            summedEventsSubmissionQueue ??
+            this.createSummedEventSubmissionQueue(metricsOptions.metricsSubmissionUrl as string, applicationInfo);
+        if (!summedEventsSubmissionQueue) {
+            return undefined;
+        }
```
perf2711, line 61: Maybe it's worth to give back some information about why metrics were not created?

packages/sdk-core/src/modules/metrics/MetricsQueue.ts
```diff
@@ -0,0 +1,7 @@
+export interface MetricsQueue<T> {
+    total: number;
+    submissionUrl: string;
+    maximumEvents: number;
```
perf2711, line 4: Should these be `readonly`?

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
```diff
@@ -0,0 +1,82 @@
...
+export class MetricsSubmissionQueue<T extends MetricsEvent> implements MetricsQueue<T> {
+    public get total() {
+        return this._events.length;
+    }
+
+    public get submissionUrl() {
+        return this._submissionUrl;
+    }
+
+    public readonly TIMEOUT_BETWEEN_REQUESTS = TimeHelper.convertSecondsToMilliseconds(10);
```
perf2711, line 16: I think you meant "delay", not "timeout".

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
```diff
@@ -0,0 +1,82 @@
...
+        );
+        if (response.status !== 'Ok') {
+            this._numberOfDroppedRequests++;
+            attempts += 1;
+            await Delay.wait(2 ** attempts * this.TIMEOUT_BETWEEN_REQUESTS);
+            await this.submitMetricsToServer(events, attempts);
+            return;
+        }
+
+        this._numberOfDroppedRequests = 0;
```
perf2711, line 67: Can we do this iteratively, rather than recursively?

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
```diff
@@ -0,0 +1,82 @@
...
+        this._events.push(event);
+        if (this.reachedLimit()) {
+            this.send();
+        }
+    }
+
+    public async send() {
+        const eventsToProcess = this._events.splice(0);
+        return await this.submitMetricsToServer(eventsToProcess);
+    }
```
perf2711, line 41: Is there a chance that the `send` method is executed before the previous send completed? Maybe we should limit this somehow?

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
```diff
@@ -0,0 +1,82 @@
...
+        this._numberOfDroppedRequests = 0;
+    }
+
+    private returnEventsIfPossible(events: T[]) {
+        if (this.maximumEvents < this._events.length + events.length) {
+            return;
+        }
+
+        // push events to the beginning of the queue
+        this._events = events.concat(this._events);
```
perf2711, line 76: You can use `this._events.unshift(...events)`.

packages/browser/src/BacktraceBrowserSessionProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.newSession = false;
+        this.sessionId = localStorage.getItem(this.SESSION_GUID) as string;
+    }
+
+    public afterMetricsSubmission(): void {
+        this.updateLastActiveTimestamp();
+    }
+
+    public shouldSend(): boolean {
+        return document.hidden === false;
```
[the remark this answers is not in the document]
konraddysput: good idea. Thanks

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -55,27 +66,33 @@ export abstract class BacktraceCoreClient {
...
-    private readonly _sessionId: string = IdGenerator.uuid();
+    private readonly _metrics?: BacktraceMetrics;
 
     protected constructor(
         protected readonly options: BacktraceConfiguration,
         private readonly _sdkOptions: SdkOptions,
         requestHandler: BacktraceRequestHandler,
         attributeProviders: BacktraceAttributeProvider[] = [],
         stackTraceConverter: BacktraceStackTraceConverter = new V8StackTraceConverter(),
+        private sessionProvider: BacktraceSessionProvider = new SingleSessionProvider(),
```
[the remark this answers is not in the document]
konraddysput: done. thx

packages/sdk-core/src/common/DelayHelper.ts
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * Promise set timeout wrapper.
+     * @param timeout timeout in ms
+     * @param signal abort signal
+     */
+    public static wait(timeout: number, signal?: AbortSignal) {
+        return new Promise<void>((resolve) => {
+            function abortCallback() {
+                clearTimeout(intervalId);
+                resolve();
```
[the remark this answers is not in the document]
konraddysput: reject requires from you to capture exceptions from wait when we have a signal handler. In most cases we want to do it when we close the client. 

reject will add more headache that we don't want. For example cancelling all requests on application exit might throw an exception or will require from us to handle this exception somehow. 

I would prefer to stay with resolve. If the team is not ok with that, lets explore options

packages/sdk-core/src/model/configuration/BacktraceConfiguration.ts
```diff
@@ -1,15 +1,27 @@
 import { BacktraceAttachment } from '../attachment';
 import { BacktraceDatabaseConfiguration } from './BacktraceDatabaseConfiguration';
 
-export interface BacktraceMetricsSupport {
+export interface BacktraceMetricsOptions {
+    /**
+     * Metrics server hostname. By default the value is set to: https://events.backtrace.io.
```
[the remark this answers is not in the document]
konraddysput: changed

packages/sdk-core/src/model/configuration/BacktraceConfiguration.ts
```diff
@@ -1,15 +1,27 @@
...
-    enable: boolean;
-    ignoreSslCertificate?: boolean;
     /**
-     * Indicates how often crash free metrics are sent to Backtrace.
+     * Determines if the metrics support is enabled. By default the value is set to true.
+     */
+    enable?: boolean;
+    /**
+     * Indicates how often crash free metrics are sent to Backtrace. The interval is a value in ms.
      * By default, session events are sent on application startup/finish, and every 30 minutes while the game is running.
```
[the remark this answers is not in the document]
konraddysput: good catch

packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public start() {
+        if (!this._sessionProvider.newSession) {
+            return;
+        }
+
+        const { attributes } = this._attributeManager.get();
+        this._summedEventsSubmissionQueue.add(
+            new SummedEvent('Application Launches', this.convertAttributes(attributes)),
+        );
```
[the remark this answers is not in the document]
konraddysput: yeah, it's a better idea. Changed.

packages/sdk-core/src/modules/metrics/MetricsBuilder.ts
konraddysput: It makes a lot of sense but the SDK right now doesn't have any logger or any logger options. Can we handle it in the future when we will be working on options related to logging?

packages/sdk-core/src/modules/metrics/MetricsQueue.ts
konraddysput: yeah, they should. Changed

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
konraddysput: Yes, it could happen. In the worst-case scenario if there are more events in the queue than maximumEvents we can lose some events. Unfortunately, there is no perfect way to do this. We can also make the interface asynchronous, but one of the requirements we got for metrics is to send them when we reach the limit. There is a tradeoff I agree. 

let me back to you with some ideas how we can change it

packages/sdk-core/src/modules/metrics/MetricsBuilder.ts
perf2711: sure

packages/sdk-core/src/common/DelayHelper.ts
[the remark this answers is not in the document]
perf2711: So from what I see, you use this to retry, without the abort callback. If you would ever use abort there, this could mean that after cancelling the delay, sending would just continue without waiting. I'm not sure whether this is what you want.

packages/browser/src/BacktraceBrowserSessionProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!window.localStorage) {
+            return;
+        }
+
+        const lastActive = this.readLastActiveTimestamp();
+        if (!lastActive || TimeHelper.now() - lastActive > this.PERSISTENCE_INTERVAL) {
+            this.updateLastActiveTimestamp();
+            localStorage.setItem(this.SESSION_GUID, this.sessionId);
+            return;
+        }
```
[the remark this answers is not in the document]
konraddysput: This information needs to exist asap in the storage, otherwise, when someone opens another tab quickly, the application will have two sessions. This is why I decided to add it to the constructor. I agree it might be weird, however, this is a real situation. 

We can move it to the store method if you would prefer, but the store method will be invoked just after the session provider creation.

I will move it to the start method that we start to use once after client initialization

packages/sdk-core/src/common/DelayHelper.ts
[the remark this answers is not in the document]
konraddysput: Let's simplify the logic for now. We don't need abort controller for now anywhere, so I will drop it from the class and we can bring it once again when we will need to start using it

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
konraddysput: The algorithm is clear and simply enough to keep it in a recursive way. I would prefer to stay with the recursively algorithm.

packages/sdk-core/src/common/DelayHelper.ts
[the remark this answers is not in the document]
konraddysput: After thinking about it once again I'm changing it to reject instead

packages/browser/tests/metrics/persistentSessionProviderTests.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    it('Should generate a new sessionId if the lastActive timestamp is greater than persistence interval time', () => {
+        const fakeId = 'test';
+        const lastSessionActiveDate = new Date(Date.now() - BacktraceBrowserSessionProvider.PERSISTENCE_INTERVAL - 1);
+        localStorage.setItem('backtrace-last-active', lastSessionActiveDate.getTime().toString(10));
+        localStorage.setItem('backtrace-guid', fakeId);
+
+        const sessionProvider = new BacktraceBrowserSessionProvider();
+        expect(sessionProvider.sessionId).toBeDefined();
+        expect(sessionProvider.sessionId).not.toEqual(fakeId);
+    });
```
perf2711, line 25: I would also do an "anti-test" of that. So something like "Should not generate a new sessionId if the lastActive timestamp is less than persistence interval time"

perf2711 (approved):

PR #37, opened 2023-07-03
Add ZIP upload of sourcemaps.

perf2711:
This diff changes uploading of sourcemaps one-by-one to a ZIP upload of all sourcemaps at once.

Symbolication ID is no longer required, as backend will read debug IDs from the sourcemaps themselves.


tools/sourcemap-tools/src/ZipArchive.ts
```diff
@@ -0,0 +1,35 @@
...
+        return this;
+    }
+
+    public finalize() {
+        return this._archive.finalize();
+    }
+
+    public override pipe<T extends NodeJS.WritableStream>(
+        destination: T,
+        options?: { end?: boolean | undefined } | undefined,
```
adamcronin42, line 23: Since `end` is already optional, do you need the `| undefined`?

tools/sourcemap-tools/src/ZipArchive.ts
```diff
@@ -0,0 +1,35 @@
...
+    private readonly _archive: archiver.Archiver;
+
+    constructor(opts?: TransformOptions) {
+        super(opts);
+        this._archive = archiver('zip');
+    }
+
+    public append(name: string, sourceMap: string | Readable | Buffer) {
+        this._archive.append(sourceMap, { name });
+        return this;
```
adamcronin42, line 14: Out of curiosity, why do we need to return `this`?

tools/webpack-plugin/src/BacktracePlugin.ts
```diff
@@ -52,45 +51,60 @@ export class BacktracePlugin implements WebpackPluginInstance {
...
+            if (this._sourceMapUploader) {
+                logger.time(`upload sourcemaps`);
                 try {
-                    const result = await this._sourceMapUploader.upload(sourceMapPath, debugId);
-                    stats.sourceMapUpload = result;
-                    logger.log(`[${asset}] file processed and sourcemap uploaded`);
+                    const archive = new ZipArchive();
+                    const request = this._sourceMapUploader.uploadSymbol(archive);
+
+                    for (const [asset, , sourceMapPath] of entries) {
```
adamcronin42, line 75: Is this syntax just ignoring the second item of entries? Could we use an underscore or something?

adamcronin42 (approved):
Looks good! Left a few small questions

tools/sourcemap-tools/src/ZipArchive.ts
perf2711: We do not, that was automatically inserted by VS Code. Will fix

tools/webpack-plugin/src/BacktracePlugin.ts
perf2711: Yeah, exactly for that. The underscore isn't really necessary there and it generates some ESLint warnings:
image

tools/sourcemap-tools/src/ZipArchive.ts
adamcronin42: Good to know!

PR #38, opened 2023-07-10
Adding React Backtrace Client, ErrorBoundary, and tests

adamcronin42:
Adding:

- A React BacktraceClient
- Adding a custom Backtrace ErrorBoundary that sends a report to Backtrace on a rendering error
- Tests

You will also see that I added an initialize() function to the React BacktraceClient. I needed this to add the client to the window but I also propose that we add an initialize function to all of our BacktraceClients as I find it to be a good user experience and not require the user to know about implementation details.

### DIFF
--- a/packages/react/src/BacktraceClient.ts
+++ b/packages/react/src/BacktraceClient.ts
@@ -9,7 +9,7 @@ import { BacktraceConfiguration } from '@backtrace/browser';
 import { BacktraceClientBuilder } from '@backtrace/browser';
 
 export class BacktraceClient extends BacktraceCoreClient {
-    private static _instance: BacktraceClient;
+    private static _instance?: BacktraceClient;
     constructor(
         options: BacktraceConfiguration,
         handler: BacktraceRequestHandler,
--- a/packages/react/src/BacktraceClient.ts
+++ b/packages/react/src/BacktraceClient.ts
@@ -1,0 +1,38 @@
+import {
+    BacktraceAttributeProvider,
+    BacktraceCoreClient,
+    BacktraceRequestHandler,
+    BacktraceStackTraceConverter,
+} from '@backtrace/sdk-core';
+import { AGENT } from '@backtrace/browser';
+import { BacktraceConfiguration } from '@backtrace/browser';
+import { BacktraceClientBuilder } from '@backtrace/browser';
+
+declare global {
+    interface Window {
+        backtraceClient?: BacktraceClient;
+    }
+}
+
+export class BacktraceClient extends BacktraceCoreClient {
+    constructor(
+        options: BacktraceConfiguration,
+        handler: BacktraceRequestHandler,
+        attributeProviders: BacktraceAttributeProvider[],
+        stackTraceConverter: BacktraceStackTraceConverter,
+    ) {
+        super(options, AGENT, handler, attributeProviders, stackTraceConverter);
+    }
+
+    public static builder(options: BacktraceConfiguration): BacktraceClientBuilder {
+        return new BacktraceClientBuilder(options);
+    }
+
+    public static initialize(options: BacktraceConfiguration): BacktraceClient {
+        const client = this.builder(options).build();
+        if (typeof window === 'object' && window.Math === Math) {
+            window.backtraceClient = client;
+        }
+        return client;
+    }
+}
--- a/packages/react/src/BacktraceClient.ts
+++ b/packages/react/src/BacktraceClient.ts
@@ -8,13 +8,8 @@ import { AGENT } from '@backtrace/browser';
 import { BacktraceConfiguration } from '@backtrace/browser';
 import { BacktraceClientBuilder } from '@backtrace/browser';
 
-declare global {
-    interface Window {
-        backtraceClient?: BacktraceClient;
-    }
-}
-
 export class BacktraceClient extends BacktraceCoreClient {
+    public static instance: BacktraceClient;
     constructor(
         options: BacktraceConfiguration,
         handler: BacktraceRequestHandler,
@@ -29,10 +24,7 @@ export class BacktraceClient extends BacktraceCoreClient {
     }
 
     public static initialize(options: BacktraceConfiguration): BacktraceClient {
-        const client = this.builder(options).build();
-        if (typeof window === 'object' && window.Math === Math) {
-            window.backtraceClient = client;
-        }
-        return client;
+        this.instance = this.builder(options).build();
+        return this.instance;
     }
 }
--- a/packages/react/src/BacktraceClient.ts
+++ b/packages/react/src/BacktraceClient.ts
@@ -9,7 +9,7 @@ import { BacktraceConfiguration } from '@backtrace/browser';
 import { BacktraceClientBuilder } from '@backtrace/browser';
 
 export class BacktraceClient extends BacktraceCoreClient {
-    public static instance: BacktraceClient;
+    private static _instance: BacktraceClient;
     constructor(
         options: BacktraceConfiguration,
         handler: BacktraceRequestHandler,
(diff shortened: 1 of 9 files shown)


packages/react/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,52 @@
+import { Component, ErrorInfo, ReactElement, ReactNode, isValidElement } from 'react';
+
+type RenderFallback = () => ReactElement;
+
+export type Props = {
+    children: ReactNode;
+    fallback?: ReactElement | RenderFallback;
+};
```
perf2711, line 8: Do you think we should allow passing `BacktraceClient`'s instance as a prop?

packages/react/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export type State = {
+    hasError: boolean;
+    error?: Error;
+};
```
perf2711, line 13: Can we use `interface` instead of `type` here and on `Props`?

packages/react/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,52 @@
...
+        super(props);
+        this.state = {
+            hasError: false,
+            error: undefined,
+        };
+    }
+
+    static getDerivedStateFromError(error: Error) {
+        return { hasError: true, error };
+    }
```
perf2711, line 26: is this used anywhere?

packages/react/src/BacktraceClient.ts
```diff
@@ -0,0 +1,38 @@
...
+} from '@backtrace/sdk-core';
+import { AGENT } from '@backtrace/browser';
+import { BacktraceConfiguration } from '@backtrace/browser';
+import { BacktraceClientBuilder } from '@backtrace/browser';
+
+declare global {
+    interface Window {
+        backtraceClient?: BacktraceClient;
+    }
+}
```
perf2711, line 15: Instead of making a global variable on `window`, maybe we could make `BacktraceClient` have a static `instance` variable?

perf2711 (commented):
Looks mostly fine to me!

Do we catch other errors (like hook errors) using this?

packages/react/src/ErrorBoundary.tsx
adamcronin42: Sure thing. I'll fix that. Any good rule of thumb you use for deciding between type and interface?

packages/react/src/ErrorBoundary.tsx
adamcronin42: This is a lifecycle method that gets invoked after an error has been thrown by a component within the error boundary. https://legacy.reactjs.org/docs/react-component.html#static-getderivedstatefromerror

packages/react/src/BacktraceClient.ts
adamcronin42: Sure thing. I'll push that up after lunch.

adamcronin42: Fixed, let me know what you think

packages/react/src/ErrorBoundary.tsx
adamcronin42: Fixed

melekr (approved):

packages/react/src/ErrorBoundary.tsx
perf2711: I use a type when I want to make an alias or a union, like:
```
type X = string;
type Y = string | number | OtherType | EvenOtherType;
```

and an interface when I define an object contract:
```
interface Z {
  field1: number;
  field2: boolean;
}
```

packages/react/src/ErrorBoundary.tsx
perf2711: Let's leave it that way until required then.

packages/react/src/BacktraceClient.ts
perf2711: The instance can be `undefined` here, right? I'm not sure why Typescript doesn't raise an error here:
```
public static instance: BacktraceClient;
```

I was thinking, maybe change this to a getter?

```
private static _instance: BacktraceClient;

public static get instance(): BacktraceClient {
  if (!this._instance) {
    throw new Error('BacktraceClient is uninitialized. Call "BacktraceClient.initialize" function first.');
  }
}
```

That way, we have more robust error handling instead of cryptic `TypeError`s.

adamcronin42: I like it! Updated.

packages/react/src/ErrorBoundary.tsx
adamcronin42: Thanks!

packages/react/src/BacktraceClient.ts
```diff
@@ -9,7 +9,7 @@ import { BacktraceConfiguration } from '@backtrace/browser';
 import { BacktraceClientBuilder } from '@backtrace/browser';
 
 export class BacktraceClient extends BacktraceCoreClient {
-    public static instance: BacktraceClient;
+    private static _instance: BacktraceClient;
```
perf2711, line 12: Can we only mark this as optional?

```
private static _instance?: BacktraceClient;
```

I know that I've written this in my suggestion without the question mark, I'm sorry.

perf2711 (approved):

packages/react/src/BacktraceClient.ts
adamcronin42: Absolutely! Done.

packages/react/src/BacktraceClient.ts
```diff
@@ -0,0 +1,37 @@
...
+        stackTraceConverter: BacktraceStackTraceConverter,
+    ) {
+        super(options, AGENT, handler, attributeProviders, stackTraceConverter);
+    }
+
+    public static builder(options: BacktraceConfiguration): BacktraceClientBuilder {
+        return new BacktraceClientBuilder(options);
+    }
+
+    public static initialize(options: BacktraceConfiguration): BacktraceClient {
```
konraddysput, line 26: Why do we need to use the initialize method? Could you take a look at you're calling builder to build the class and store it in the class instance of the client? You can do the same in the constructor, which is a more intuitive way for me. 

WIth this approach, you have one way to initialize everything. Right now, we have a different way to initialize other clients and we don't know which one (builder vs client) is preferred.

packages/react/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,52 @@
...
+        };
+        // grabbing here so it will fail fast if BacktraceClient is uninitialized
+        this._client = BacktraceClient.instance;
+    }
+
+    static getDerivedStateFromError(error: Error) {
+        return { hasError: true, error };
+    }
+
+    componentDidCatch(error: Error, info: ErrorInfo) {
```
konraddysput, line 32: shouldn't we also add accessors like public or private? We use it everywhere in the code. Is there any reason why we shouldn't continue doing that here?

adamcronin42: Good call I can add them

packages/react/src/BacktraceClient.ts
adamcronin42: I propose we add an initialize method to all of our clients that calls the builder. I think it would provide a good user experience as the user would not be required to know implementation details about which patterns we use to create the client.

perf2711: In some way I agree with @adamcronin42 that `.initialize` may be a better method. Or the one that actually initializes Backtrace in a global context. `.build` could only create the instance, and not assign it as a global error handler.

packages/react/src/ErrorBoundary.tsx
adamcronin42: Fixed in latest PR

PR #40, opened 2023-07-11
Add CLI for manipulating source and sourcemap files.

perf2711:
This PR adds CLI with three commands:
* `add-sources` - adds sources from original source files to the sourcemap,
* `process` - adds debug IDs to sources and sourcemaps,
* `upload` - uploads sourcemaps to Backtrace.


tools/cli/src/logger.ts
```diff
@@ -0,0 +1,112 @@
+import { LogLevel, Logger } from '@backtrace/sourcemap-tools';
+import { format } from 'util';
+
+export interface LoggerOptions {
+    readonly verbose?: boolean[];
+    readonly quiet?: boolean;
+    readonly 'log-level'?: CliLogLevel;
```
adamcronin42, line 7: Out of curiosity, why not just use camelCase?

tools/cli/src/sourcemaps/add-sources.ts
```diff
@@ -0,0 +1,118 @@
...
+        name: 'path',
+        description: 'Path to sourcemap files to append sources to.',
+        defaultOption: true,
+        defaultValue: process.cwd(),
+        multiple: true,
+        alias: 'p',
+    })
+    .option({
+        name: 'dry-run',
+        alias: 'n',
```
adamcronin42, line 32: Is 'n' an existing convention? What about 'd'?

adamcronin42 (approved):

tools/cli/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        description: 'Verbosity level. -v prints info logs, -vv prints debug logs, -vvv prints ALL logs.',
+    })
+    .option({
+        name: 'quiet',
+        type: Boolean,
+        alias: 'q',
+        global: true,
+        description: 'Disables ALL logging messages.',
+    })
+    .option({
```
konraddysput, line 40: What is default? Can we include what is default behavior in the log message?

tools/cli/src/commands/Command.ts
```diff
@@ -0,0 +1,166 @@
+import { Err, Ok, Result } from '@backtrace/sourcemap-tools';
+import commandLineArgs from 'command-line-args';
+import commandLineUsage, { Section } from 'command-line-usage';
+import { LoggerOptions, createLogger } from '../logger';
+import { CommandError } from '../models/CommandError';
+import { ExtendedOptionDefinition } from '../models/OptionDefinition';
+
+export class Command<T extends object = any> {
```
konraddysput, line 8: Does it need to be any? We know what type we expect here. Does it need to be generic?

tools/cli/src/commands/Command.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    public getHelpMessage(stack?: Command[]) {
+        const globalOptions = [
+            ...this.options.filter((o) => o.global),
+            ...(stack?.flatMap((o) => o.options.filter((o) => o.global)) ?? []),
+        ];
+
+        const nonGlobalOptions = this.options.filter((o) => !o.global);
+
+        let cmd = 'backtrace';
```
konraddysput, line 98: can we expose prefix to const? I think we want to have it in const in case we want to change it for a different command. `Backtrace` sounds weird because you won't use it to other operations you can do in Backtrace.

tools/cli/src/helpers/find.ts
```diff
@@ -0,0 +1,35 @@
+import { Err, FileFinder, Ok, ResultPromise } from '@backtrace/sourcemap-tools';
+import fs from 'fs';
+import path from 'path';
+
+export async function find(regex: RegExp, ...paths: string[]): ResultPromise<string[], string> {
```
konraddysput, line 5: hmm can we be more descriptive what we want to find here?

konraddysput (approved):
Added a few comments to be more descriptive, but overall the diff looks really good. No required changes on my side

tools/cli/src/logger.ts
perf2711: I wanted to do args in POSIX commands style, and I've never seen a command with camelCase switches.

For example, exempt from `man rsync`:
```
-L, --copy-links            transform symlink into referent file/dir
--copy-unsafe-links     only "unsafe" symlinks are transformed
--safe-links            ignore symlinks that point outside the tree
-k, --copy-dirlinks         transform symlink to dir into referent dir
-K, --keep-dirlinks         treat symlinked dir on receiver as dir
```

tools/cli/src/sourcemaps/add-sources.ts
perf2711: I believe so. Most places I've seen have `-n` alias as the dry run switch.

E.g. `rsync`:
```
-n, --dry-run               show what would have been transferred
```

tools/cli/package.json
```diff
@@ -0,0 +1,48 @@
+{
+    "name": "@backtrace/javascript-cli",
```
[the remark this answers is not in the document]
perf2711: Yes, add:
```
{
  "bin": {
    "<execName>": "lib/index.js"
  }
}
```

tools/cli/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+})
+    .subcommand(processCmd)
+    .subcommand(uploadCmd)
+    .subcommand(addSourcesCmd)
+    .option({
+        name: 'help',
+        type: Boolean,
+        alias: 'h',
+        global: true,
+        description: 'Displays this help message.',
```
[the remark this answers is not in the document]
perf2711: It makes more sense when you see the help message:
```
backtrace

  Backtrace utility for managing Javascript files. 

  Usage: backtrace <command> [options] 

Available commands

  process       Processing source and sourcemap files 
  upload        Uploading of sourcemaps to Backtrace  
  add-sources   Adds sources to sourcemap files       

Global options

  -h, --help           Displays this help message.                              
  -v, --verbose        Verbosity level. -v prints info logs, -vv prints debug   
                       logs, -vvv prints ALL logs.                              
  -q, --quiet          Disables ALL logging messages.                           
  --log-level string   Sets the logging level. Can be one of: quiet, error,     
                       warn, info, debug, verbose
```

So `--help` is referring to _this_ help message.

tools/cli/src/commands/Command.ts
perf2711: Done.

tools/cli/src/index.ts
perf2711: Added a default description.

PR #43, opened 2023-07-18
Source debug ID insertion fixes.

perf2711:
This PR fixes two issues:

* When the source has a shebang on the first line, the debug ID snippet was inserted before it and not after it
* When the source had trailing whitespaces, the debug ID comment was inserted after them, and not before them


tools/sourcemap-tools/src/helpers/stringHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+ * const value = 'def';
+ * const appended = appendBeforeWhitespaces(str, value);
+ */
+export function appendBeforeWhitespaces(str: string, value: string) {
+    const whitespaces = str.match(/\s*$/)?.[0];
+    if (!whitespaces) {
+        return str + value;
+    }
+
+    return str.substring(0, whitespaces.length) + value + whitespaces;
```
adamcronin42, line 16: I may not be understanding this correctly, but do we want to use `whitespaces.length` here? Wouldn't we want str without the whitespaces + value + whitespaces?

perf2711: Yeah, I screwed that up. Thanks for the heads-up.

I've added some unit tests.

adamcronin42 (approved):
Looks good!

konraddysput (approved):

tools/sourcemap-tools/src/helpers/stringHelpers.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Appends `value` to `str` before trailing whitespaces in `str`.
+ * @param str String to append to.
+ * @param value String to append.
+ * @example
+ * const str = 'abc\n\n';
+ * const value = 'def';
+ * const appended = appendBeforeWhitespaces(str, value);
```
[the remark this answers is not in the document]
perf2711: done

PR #44, opened 2023-07-18
Feature/react example app

adamcronin42:
This is an example React v18 app that has a button to trigger the Backtrace ErrorBoundary and also a button to manually send a report to Backtrace.

image


perf2711 (approved):

examples/sdk/react/README.md
```diff
@@ -0,0 +1,64 @@
+# Getting Started with Create React App
```
konraddysput, line 1: This shouldn't be here. Can we please adjust readme to our example?

examples/sdk/react/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "@backtrace/react-example",
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "@testing-library/jest-dom": "^5.16.5",
+        "@testing-library/react": "^13.4.0",
+        "@testing-library/user-event": "^13.5.0",
+        "@types/jest": "^27.5.2",
+        "@types/node": "^16.18.37",
```
konraddysput, line 10: types/node?

examples/sdk/react/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "@backtrace/react-example",
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "@testing-library/jest-dom": "^5.16.5",
```
konraddysput, line 6: @testing libraries shouldn't be in the dev dependencies list?

examples/sdk/react/src/App.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { useState } from 'react';
+import './App.css';
+import { BacktraceClient } from '@backtrace/react';
+
+function App() {
+    const [clicked, setClicked] = useState(false);
+    function sendError() {
+        const client = BacktraceClient.instance;
+        if (client) {
```
konraddysput, line 9: I think it's important to make it loud the client is not available for any reason. Can we throw an error?

examples/sdk/react/src/App.tsx
```diff
@@ -0,0 +1,75 @@
...
+            client.send(new Error('Manual Test Error'));
+            alert('Error sent! Check your Backtrace console to see the error.');
```
konraddysput, line 11: the method is asynchronous. Can we make it async? I think it would be great if we can make it sync later but you can't expect to synchronously send it to the server.

Alert is also ... easy way to demonstrate something, but not a really cool way to do that. Wouldn't it be better to use toast/div error message instead?

examples/sdk/react/src/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import ReactDOM from 'react-dom/client';
+import App from './App';
+import { ErrorBoundary, BacktraceClient } from '@backtrace/react';
+import { Fallback } from './Fallback';
+
+const universe = 'your-universe';
```
konraddysput, line 7: Can we make all our examples similar? can we expose it to consts like in the rest of our example and just store the submission URL? It's better to have only submission URL because this is what people will see in the integration guide

konraddysput (changes requested):
1. Can we please adjust consts like submission URL?
2. Can we please adjust README.md? 
3. Can we add more examples - message/error report & metrics ?

examples/sdk/react/src/App.tsx
adamcronin42: We throw an error in the instance getter if the client is not available.

examples/sdk/react/package.json
adamcronin42: This and all of the testing library stuff are just part of the default CRA typescript template. I've removed them now.

examples/sdk/react/src/App.tsx
adamcronin42: async -> fixed

Also I added a toast when the error gets send and also if the submission url hasn't been updated.

examples/sdk/react/src/index.tsx
adamcronin42: Fixed

examples/sdk/react/README.md
adamcronin42: Fixed

adamcronin42:
> Can we please adjust consts like submission URL?
> Can we please adjust README.md?
> Can we add more examples - message/error report & metrics ?

@konraddysput all fixed. The one issue I ran into is when I tried to add metrics I get 401s every time I call `client.metrics.send()` even though metrics exists. Any idea why that might be?

examples/sdk/react/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### `npm run build`
+
+Builds the app for production to the `build` folder.\
+It correctly bundles React in production mode and optimizes the build for the best performance.
+
+The build is minified and the filenames include the hashes.\
+Your app is ready to be deployed!
+
+See the section about [deployment](https://facebook.github.io/create-react-app/docs/deployment) for more information.
```
konraddysput, line 34: should this information be here?

konraddysput (approved):

examples/sdk/react/README.md
adamcronin42: Good call, fixed

PR #46, opened 2023-07-25
Add bin entry to CLI.

perf2711:
This adds `backtrace-js` as a bin entry to enable using `backtrace-js` in the shell.

### DIFF
--- a/tools/cli/package.json
+++ b/tools/cli/package.json
@@ -7,6 +7,9 @@
     "engines": {
         "node": ">=14"
     },
+    "bin": {
+        "backtrace-js": "./lib/index.js"
+    },
     "scripts": {
         "build": "tsc",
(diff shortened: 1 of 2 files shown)


tools/cli/package.json
```diff
@@ -7,6 +7,9 @@
     "engines": {
         "node": ">=14"
     },
+    "bin": {
+        "backtrace-js": "./lib/index.js"
```
konraddysput, line 11: what other functions will this library have? So far we know only about the source map support. Do we plan to add anything else?

konraddysput (approved):

tools/cli/package.json
perf2711: I don't really know about that. We could certainly extend this to have other functionalities as the need for them arises.

adamcronin42 (approved):

PR #47, opened 2023-07-26
Feature/component stack as backtracedata

adamcronin42:
### What's Included?

- Adding the ability to send BacktraceData to client.send() -> this is so we can pass the React component stack as an additional thread and get out-of-the-box support in the debugger for component stacks
- Sending the error and component stack in the ErrorBoundary.componentDidCatch lifecycle method
- Detecting React v16 component stacks
- Parsing React v16 component stacks
- tests

### What's Next?

- Naming Error Boundaries
- Passing React-specific attributes by default with reports
- Allow users to pass their own additional attributes to the ErrorBoundary
- Hook up with source map stuff


konraddysput (changes requested):
I really don't like this approach. The way how it is designed is to expose options to delivered backtrace clients to use what they need. 

For example: 
- we can make those properties protected, so backtraceClient in the react client can use it for any need
- we can creating client send definition that client by default can't use. It generates more confusion. 

Instead, my proposals are:
* allow Backtrace report to accept other threads (needed also in the react-native and electron integration soon). We can use additional backtrace reports options
* In the stack trace converter you can write your own converter and decide what to use and when. In your converter, you can detect type of the frame and use a correct parser. You can pass later the converter to the core client and use it.


By doing the following we can avoid creating solutions just for one single framework and use existing solutions that were created for this purpose. We can also keep our algorithm simple.

packages/browser/src/builder/BacktraceClientBuilder.ts
```diff
@@ -13,7 +13,7 @@ import { SpiderMonkeyStackTraceConverter } from '../converters/SpiderMonkeyStack
 import { getEngine } from '../engineDetector';
 
 export class BacktraceClientBuilder extends BacktraceCoreClientBuilder<BacktraceClient> {
-    constructor(private readonly _options: BacktraceConfiguration) {
+    constructor(protected readonly _options: BacktraceConfiguration) {
```
konraddysput, line 16:
```suggestion
    constructor(protected readonly options: BacktraceConfiguration) {
```

1. if it is protected, can we remove `_` ?
2. Do you need it for any reason?

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -47,23 +57,29 @@ export class BacktraceReport {
         if (data instanceof Error) {
             this.annotations['error'] = data;
             this.classifiers = [data.name];
-            this.message = data.message;
-            this.stackTrace = data.stack ?? '';
+            const stackInfo = {
+                stack: data.stack ?? '',
+                message: data.message,
+            };
+            this.stackTrace['main'] = stackInfo;
```
konraddysput, line 64:
```suggestion
            this.stackTrace['main'] = {
                stack: data.stack ?? '',
                message: data.message,
            };
```

what do you think about this?

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -47,23 +57,29 @@ export class BacktraceReport {
...
             errorType = 'Message';
             this.skipFrames += 1;
         }
 
         if (!this.attributes['error.type']) {
             this.attributes['error.type'] = errorType;
         }
-        this.attributes['error.message'] = this.message;
+        this.attributes['error.message'] = this.stackTrace['main']?.message ?? '';
```
konraddysput, line 83: We still need to set the error message. It's important to have an easy way to verify the report message by the user. Can we still maintain the message property and use it here?

packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
```diff
@@ -68,4 +51,32 @@ export class BacktraceDataBuilder {
...
+            };
+
+            for (const frame of stackFrames) {
+                const debugIdentifier = this._debugIdProvider.getDebugId(frame.library);
+                if (!debugIdentifier) {
+                    continue;
+                }
+                detectedDebugIdentifier = true;
+                frame.debug_identifier = debugIdentifier;
+            }
```
konraddysput, line 75: it feels more natural to set the thread once we add debug_identifiers. The current approach should still work, however it's easy to mess up something when we modify the code and we don't have the same reference to the object. Can we set the thread after `for (const frame of stackFrames) {` loop?

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -139,6 +139,12 @@ export abstract class BacktraceCoreClient {
      * @param report Backtrace Report
      */
     public async send(report: BacktraceReport): Promise<void>;
+
```
konraddysput, line 142: the comment is invalid. We don't send Backtrace data anymore. Can we remove it and still use description from each send method above?

packages/react/src/helpers/componentStackHelpers.ts
```diff
@@ -0,0 +1,29 @@
+import { BacktraceStackFrame } from '@backtrace/sdk-core/src/model/data/BacktraceStackTrace';
+
+// React 16 component stacks need to be parsed separately. React 17+ component stacks can be parsed like Error stacks
+export function isReact16ComponentStack(stack: string): boolean {
+    if (!stack) return false;
```
konraddysput, line 5:
```suggestion
    if (!stack) { 
        return false;
    }
```

can we continue using brackets? This is the pattern we have in the code.

packages/react/src/helpers/componentStackHelpers.ts
```diff
@@ -0,0 +1,29 @@
+import { BacktraceStackFrame } from '@backtrace/sdk-core/src/model/data/BacktraceStackTrace';
+
+// React 16 component stacks need to be parsed separately. React 17+ component stacks can be parsed like Error stacks
```
konraddysput, line 3: can we use /** ?

packages/react/src/helpers/componentStackHelpers.ts
```diff
@@ -0,0 +1,29 @@
...
+        if (!frame.includes('in ')) {
+            return false;
+        }
+    }
+    return true;
+}
+
+export function parseReact16ComponentStack(stack: string): BacktraceStackFrame[] {
+    const btFrames: BacktraceStackFrame[] = [];
+    if (!stack) return btFrames;
```
konraddysput, line 17:
```suggestion
    if (!stack){
        return btFrames;
    }
```

packages/react/src/helpers/componentStackHelpers.ts
```diff
@@ -0,0 +1,29 @@
...
+    const frames = stack.split('\n');
+    for (const frame of frames) {
+        if (!frame.trim()) continue;
+        const component = frame.split('in ')?.[1]?.trim();
```
konraddysput, line 21: split will always return an array. Do we need to add `?.`

packages/react/src/converters/ReactStackTraceConverter.ts
```diff
@@ -0,0 +1,20 @@
+import { BacktraceStackTraceConverter } from '@backtrace/sdk-core';
+import { JavaScriptEngine } from '@backtrace/sdk-core/src/model/data/JavaScriptEngine';
+import { BacktraceStackFrame } from '@backtrace/sdk-core/lib/model/data/BacktraceStackTrace';
+import { isReact16ComponentStack, parseReact16ComponentStack } from '../helpers/componentStackHelpers';
+
+export class ReactStackTraceConverter implements BacktraceStackTraceConverter {
```
konraddysput, line 6: how about moving everything from compontentStackHelper here?  I think it does make sense to keep simple and small logic inside one dedicated file.

konraddysput (changes requested):
Hey,

After revisiting the diff once again I think we need to make at least one more change before we land it - I believe it's important to maintain the report message and still use it. Based on this users might want to decide to filter/modify something during the send stage.

packages/browser/src/builder/BacktraceClientBuilder.ts
adamcronin42: 1. Fixed
2. I do need it when creating a BacktraceClient in `BacktraceReactClientBuilder.build()`

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -36,6 +32,20 @@ export class BacktraceReport {
...
     public skipFrames = 0;
 
+    /**
+     * Add additional stack trace to the report.
+     * If the thread name already exists it will be overwritten
+     * @param name thread name
+     * @param stack stack trace
+     * @param message error message
+     */
+    public addStackTrace(name: string, stack: string, message: string) {
```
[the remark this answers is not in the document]
adamcronin42: Good point, fixed

packages/sdk-core/src/model/report/BacktraceReport.ts
adamcronin42: I like it, fixed

packages/sdk-core/src/model/report/BacktraceReport.ts
adamcronin42: I added message back as a field in BacktraceReport. I set it as well as the message in the main stack trace

packages/sdk-core/src/modules/data/BacktraceDataBuilder.ts
adamcronin42: Fair point, fixed

packages/sdk-core/src/BacktraceCoreClient.ts
adamcronin42: Good catch, I removed the comment

packages/react/src/helpers/componentStackHelpers.ts
adamcronin42: Fixed

packages/react/src/helpers/componentStackHelpers.ts
adamcronin42: Fixed

packages/react/src/converters/ReactStackTraceConverter.ts
adamcronin42: I moved everything over and also updated the tests

packages/react/src/converters/ReactStackTraceConverter.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    public convert(stackTrace: string, message = ''): BacktraceStackFrame[] {
+        // React 16 component stacks are not JS error stacks, and need to be parsed separately
+        if (this.isReact16ComponentStack(stackTrace)) {
+            return this.parseReact16ComponentStack(stackTrace);
+        }
+        return this.stackTraceConverter.convert(stackTrace, message);
+    }
+
+    /** React 16 component stacks need to be parsed separately. React 17+ component stacks can be parsed like Error stacks */
```
konraddysput, line 20: the comment format is different. Can you use
```
/**
*
*/
```

konraddysput (approved):

perf2711 (approved):

packages/react/src/converters/ReactStackTraceConverter.ts
adamcronin42: Fixed

PR #49, opened 2023-07-27
Before send event support

konraddysput:
# Why

This diff adds before send event to the client. We're trying to be consistent between different clients and SDKs that we offer. All our managed SDKs offer "before send" event to modify data or even skip sending it for any reason. Why do customers want to use it?
- based on dynamic criteria, decide to not send it
- apply PII settings 
- apply custom attribute logic when error happen

### DIFF
--- a/packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
+++ b/packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
@@ -1,0 +1,50 @@
+import { BacktraceReportSubmissionResult } from '../../src';
+import { BacktraceData } from '../../src/model/data/BacktraceData';
+import { BacktraceTestClient } from '../mocks/BacktraceTestClient';
+
+describe('Client callbacks tests', () => {
+    describe('Submission data modification tests', () => {
+        it('Should invoke the before send event', async () => {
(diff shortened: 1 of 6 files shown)


perf2711 (approved):

packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        it('Should skip sending events if the user decide not to send it', async () => {
+            const client = BacktraceTestClient.buildFakeClient({
+                beforeSend: () => {
+                    return undefined;
+                },
+            });
+
+            await client.send(new Error());
+            expect(client.requestHandler.postError).not.toHaveBeenCalled();
```
adamcronin42, line 47: Do you not have to spy on this to see if it was called?

adamcronin42 (approved):
Looks good! Left one question.

packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
konraddysput: It's alerady in wrapped by using spy in the test backtrace client implementation

PR #50, opened 2023-07-27
Skip report callback

konraddysput:
# Why

This diff adds support for skip report callback. Users should use it to decide if the report should be filtered or not before we generate diagnostic information.

### DIFF
--- a/packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
+++ b/packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
@@ -67,7 +67,7 @@ describe('Client callbacks tests', () => {
             expect(client.requestHandler.postError).toHaveBeenCalled();
         });
 
-        it('Should send a report if the filter returns true ', async () => {
+        it('Should not send a report if the filter returns true ', async () => {
             const client = BacktraceTestClient.buildFakeClient({
                 skipReport: () => {
                     return true;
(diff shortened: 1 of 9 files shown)


packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
```diff
@@ -47,4 +47,35 @@ describe('Client callbacks tests', () => {
...
+                skipReport: () => {
+                    return false;
+                },
+            });
+
+            await client.send(new Error());
+            expect(client.requestHandler.postError).toHaveBeenCalled();
+        });
+
+        it('Should send a report if the filter returns true ', async () => {
```
adamcronin42, line 70: 'Should _not_ send report..'

adamcronin42 (approved):
Looks good! Left one small comment on a typo.

packages/sdk-core/tests/client/clientCallbacksTests.spec.ts
konraddysput: ups. Thank yoU!

konraddysput:
> Do you think we could make the `skipReport` method asynchronous?

we cannot use asynchronous methods in the error reporting flow - every time we reach the asynchronous function, the unhandled exception handler won't wait on us. We can later also change the send method and not require it to be asynchronous. I will cover those details later.

PR #51, opened 2023-07-27
Browser unhandler error handler

konraddysput:
# Why

This diff adds exception handlers to the browser package. With them, we can start reporting unhandled errors via browser.


perf2711 (approved):

adamcronin42 (approved):
Nice!

PR #58, opened 2023-08-04
sourcemap tools: add stripping of sourceContent by default

perf2711:
This mostly changes how bundle plugins work. If `includeSources` is not `true`, the sources will be stripped from the sourcemap before uploading.


tools/cli/src/sourcemaps/upload.ts
```diff
@@ -149,7 +149,7 @@ export const uploadCmd = new Command<UploadOptions>({
         const createArchiveCommand = (assets: Asset[]) =>
             AsyncResult.fromValue<Asset[], string>(assets)
                 .then(logTrace('creating archive'))
-                .then((assets) => archiveSourceMaps(sourceProcessor)(assets.map((a) => a.path)))
+                .then((assets) => archiveSourceMaps(sourceProcessor)(assets as never)) // TODO: Fix this
```
perf2711, line 152: This is fixed here: https://github.com/backtrace-labs/backtrace-javascript/pull/58

konraddysput (approved):

adamcronin42 (approved):

melekr (approved):

PR #62, opened 2023-08-08
Database support

konraddysput:
# Why

This diff adds the basic database support to the Node SDK. By doing that we can store offline our reports. When it's useful? When we retrieve an error and an application restarts (unhandled exception handlers). Or when the PC is offline for any reason and later becomes online.

This diff contains:
* database business flow in the client 
* database tests

The expectation here is to make sure the database:
* store record on send
* deletes record on successful send
* after restart, the client reads and sends all records 

To avoid extending too much the scope I didn't include:
* flush
* send
* other database methods

Where do we want to use the database integration?
- react native (here we should expect a different file system provider)
- electon

This is the basic concept of the database. All comments are welcome.


packages/node/src/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -0,0 +1,98 @@
...
+    BacktraceDatabaseStorageProvider,
+} from '@backtrace/sdk-core';
+import fs from 'fs';
+import * as fsPromise from 'fs/promises';
+import path from 'path';
+import { BacktraceDatabaseFileRecord } from './BacktraceDatabaseFileRecord';
+export class BacktraceDatabaseFileStorageProvider implements BacktraceDatabaseStorageProvider {
+    private _enabled = true;
+
+    private readonly RECORD_SUFFIX = '-recod.json';
```
perf2711, line 13: typo? shouldn't it be `-record.json`?

packages/node/src/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public static createIfValid(
+        options?: BacktraceDatabaseConfiguration,
+    ): BacktraceDatabaseFileStorageProvider | undefined {
+        if (!options) {
+            return undefined;
+        }
+        if (!options.enabled || !options.path) {
+            return undefined;
+        }
```
perf2711, line 29: Why not throw on missing path if it is enabled?

packages/node/src/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -0,0 +1,98 @@
...
+    public start(): boolean {
+        // make sure by mistake we don't create anything or start any operation
+        if (this._enabled === false) {
+            return false;
+        }
+
+        let databaseDirectoryExists = fs.existsSync(this._path);
+        if (this._createDatabaseDirectory !== false && !databaseDirectoryExists) {
+            fs.mkdirSync(this._path, { recursive: true });
+            databaseDirectoryExists = true;
```
perf2711, line 43: you could just do `return true;` here and not update `databaseDirectoryExists`

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -0,0 +1,201 @@
...
+                try {
+                    record.locked = true;
+                    const result = await this._requestHandler.send(record.data, record.attachments);
+                    if (result.status === 'Ok') {
+                        this.remove(record);
+                        continue;
+                    }
+                    submissionFailure = true;
+                    this._databaseRecordContext.increaseBucket(bucketIndex);
+                    break;
```
perf2711, line 177: Instead of assigning to `submissionFailure` and checking it afterwards, you could just `return` here.

packages/sdk-core/src/modules/database/model/BacktraceDatabaseRecord.ts
```diff
@@ -0,0 +1,14 @@
+import { BacktraceAttachment } from '../../../model/attachment';
+import { BacktraceData } from '../../../model/data/BacktraceData';
+
+export interface BacktraceDatabaseRecord {
+    data: BacktraceData;
+    id: string;
+    attachments: BacktraceAttachment[];
+    count: number;
+    hash: string;
```
perf2711, line 9: can we make these `readonly`?

packages/node/src/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+    }
+
+    public async delete(record: BacktraceDatabaseRecord): Promise<boolean> {
+        const recordPath = this.getRecordPath(record.id);
+        if (!fs.existsSync(recordPath)) {
+            return false;
+        }
+
+        await fsPromise.unlink(recordPath);
```
adamcronin42, line 62: Would we want a try catch for if the delete fails?

packages/node/src/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+    }
+
+    public add(record: BacktraceDatabaseRecord): void {
+        const recordPath = this.getRecordPath(record.id);
+        fs.writeFileSync(recordPath, JSON.stringify(BacktraceDatabaseFileRecord.fromRecord(record)), {
```
adamcronin42, line 68: Would we want a try catch for writing files?

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -0,0 +1,179 @@
...
+            await this.sendRecords();
+        };
+        this._intervalId = setInterval(sendDatabaseReports, this._retryInterval);
+        await this.sendRecords();
+    }
+
+    private async sendRecords() {
+        for (let bucketIndex = 0; bucketIndex < this._databaseRecordContext.bucketCount; bucketIndex++) {
+            for (const record of this._databaseRecordContext.getBucket(bucketIndex)) {
+                if (record.locked) {
```
adamcronin42, line 161: I'm a little confused by the purpose of the locked logic. Would you mind explaining?

adamcronin42 (approved):
Very cool! Left a few questions.

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
adamcronin42: Ah that makes sense. Thank you!

PR #64, opened 2023-08-09
Browser breadcrumbs support

konraddysput:
# Why

This diff adds browser/react breadcrumbs support. Right now we support: 
- console.log breadcrumbs
- HTTP requests
- user actions
- navigation events

### DIFF
--- a/packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
+++ b/packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
@@ -14,6 +14,7 @@ export class WebRequestEventSubscriber implements BreadcrumbsEventSubscriber {
             return;
         }
         const xmlHttpRequestOriginalOpenMethod = XMLHttpRequest.prototype.open;
+
         XMLHttpRequest.prototype.open = function (
             method: string,
             url: string | URL,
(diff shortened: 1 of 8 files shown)


packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        const fetchOriginalMethod = window.fetch;
+
+        window.fetch = async (...args) => {
+            const [resource, config] = args;
+
+            const response = await fetchOriginalMethod(resource, config);
+
+            breadcrumbsManager.addBreadcrumb(
+                `Sending an HTTP ${config?.method} to request to ${resource}. Response status: ${response.status}`,
```
perf2711, line 44: The request has been already sent - it should be `Sent an HTTP ${config?.method} request to ${resource}`.

I think that we should assume that GET is default, if method is missing.

packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
```diff
@@ -0,0 +1,70 @@
...
+            xmlHttpRequestOriginalOpenMethod.call(this, method, url, async || true, username, password);
+        };
+        this._xmlHttpRequestOriginalOpenMethod = xmlHttpRequestOriginalOpenMethod;
+
+        const fetchOriginalMethod = window.fetch;
+
+        window.fetch = async (...args) => {
+            const [resource, config] = args;
+
+            const response = await fetchOriginalMethod(resource, config);
```
perf2711, line 41: Does this throw? It would be nice to add an error breadcrumb.

packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
```diff
@@ -0,0 +1,70 @@
...
+        if ((breadcrumbsManager.breadcrumbsType & BreadcrumbType.Http) !== BreadcrumbType.Http) {
+            return;
+        }
+        const xmlHttpRequestOriginalOpenMethod = XMLHttpRequest.prototype.open;
+        XMLHttpRequest.prototype.open = function (
+            method: string,
+            url: string | URL,
+            async?: boolean,
+            username?: string | null,
+            password?: string | null,
```
perf2711, line 22: You could replace these with `...args: any[]` and pass them to `call`.

packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
konraddysput: you're correct. Adjusting the code. Thank yoU!

packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
konraddysput: yes it throws and I didn't capture at error to propagate it to the code that wanted to do it, however adding a breadcrumb about the failure is a really nice idea. I will add it

packages/browser/src/breadcrumbs/WebRequestEventSubscriber.ts
konraddysput: Types complain a lot about missing the async parameter or the existence of it. Explained on slack

perf2711 (approved):

adamcronin42 (approved):
Very cool! I think a good follow up for the click events for us would be to attempt to parse the text of the element that gets clicked as it would provide a lot of context and allow the user to quickly determine which element was clicked. For simplicity at first we could even only attempt to get the text if the tag is a <button> or <a>.

adamcronin42 (approved):

PR #66, opened 2023-08-09
node example: add OOM crash example

perf2711:


examples/sdk/node/src/index.ts
```diff
@@ -63,48 +63,68 @@ function sendMetrics() {
...
+}
+
 function showMenu() {
+    const options = [
+        ['Send an exception', (attributes: Record<string, number>) => sendHandledException(attributes)],
+        ['Send a message', (attributes: Record<string, number>) => sendMessage('test message', attributes)],
+        ['Throw rejected promise', () => rejectPromise('Rejected promise')],
+        ['Add a new summed event', (attributes: Record<string, number>) => addEvent('Option clicked', attributes)],
+        ['Send all metrics', sendMetrics],
+        ['OOM', oom],
```
konraddysput, line 101: What do you think about moving OOM just below rejected promise? For me it makes more sense because we're grouping actions one by one

konraddysput (approved):

examples/sdk/node/src/index.ts
perf2711: done

PR #67, opened 2023-08-10
cli: quality of life improvements

perf2711:
* `--version` option
* Additional messages about missing files for processing/upload/adding sources
* Add `.backracejsrc` file support, which enables configuration of every command line option, for example:
  ```jsonc
  {
      // Supports comments!
      "path": "global path",
      "add-sources": {
          "path": "add-sources specific path"
      },
      "upload": {
          "url": "url for uploading"
      }
  }
  ```

### DIFF
--- a/tools/cli/src/helpers/version.ts
+++ b/tools/cli/src/helpers/version.ts
@@ -6,7 +6,11 @@ interface PackageJson {
 }
 
 export function loadVersion() {
-    return AsyncResult.equip(readFile(path.join(__dirname, '../../package.json')))
+    const packageJsonPath = require.main?.path
+        ? path.join(require.main?.path, '..', 'package.json')
+        : path.join(__dirname, '../../package.json');
+
+    return AsyncResult.equip(readFile(packageJsonPath))
         .then(parseJSON<PackageJson>)
         .then((p) => p.version).inner;
 }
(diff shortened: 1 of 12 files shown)


tools/cli/src/helpers/version.ts
```diff
@@ -0,0 +1,12 @@
+import { AsyncResult, parseJSON, readFile } from '@backtrace/sourcemap-tools';
+import path from 'path';
+
+interface PackageJson {
+    readonly version: string;
+}
+
+export function loadVersion() {
+    return AsyncResult.equip(readFile(path.join(__dirname, '../../package.json')))
```
konraddysput, line 9: The approach you took here is technically correct, however it's super easy to make a mistake with it. Let's say now we're changing for any reason the directory structure. It's a trivial change and nothing should break, right? 

with this simple change, we're introducing an error with missing package.json version. 

Instead: can we verify if:
`path.join(require.main.path, '..', 'package.json')` exists? The approach is better, because the structure of the entry point of the package shouldn't change or can change but definitely less often. 

what do you think?

konraddysput (approved):

tools/cli/src/helpers/version.ts
perf2711: done

PR #69, opened 2023-08-10
cli: add README.md

perf2711:


perf2711:
@ianrice07 I've added you as a reviewer because you've written the other doc file. Feel free to resign

rick-bt (approved):
This is great - let's merge.
We can come back later and copy edit if we want.

PR #74, opened 2023-08-15
Feature/react redux middleware

adamcronin42:
This PR adds a redux middleware to the browser package. I was originally going to add it to the react sdk but it's actually UI/framework-agnostic so I added it to browser and users should be able to reuse it with react no problem.

What's Included:

- A Backtrace redux middleware that saves breadcrumbs when actions are dispatched. It includes the whole action object.
- An interceptAction function that can optionally be passed to the middleware. The middleware will skip the action if the function returns undefined or will save a breadcrumb for whatever action the function returns (this will allow PII or other information to be removed before sending to Backtrace). It allows for specific logic for specific actions, or can apply broad logic to all actions (ex: removing email from action if it exists).
- Tests

What's Next:

- Integrating this middleware into our example browser/react applications


packages/browser/package.json
```diff
@@ -37,18 +37,19 @@
         "/lib"
     ],
     "devDependencies": {
+        "@reduxjs/toolkit": "^1.9.5",
```
perf2711, line 40: I would add `@reduxjs/toolkit` as a `peerDependency` and externalize it in Webpack. We should not include it with our package.

packages/browser/src/redux/BacktraceReduxMiddleware.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BacktraceClient } from '../BacktraceClient';
+
+/**
+ *
+ * @param client BacktraceClient used to send breadcrumbs
+ * @param interceptAction A function that can be used to skip an action or filter out information from a dispatched action, such as PII, that shouldn't be sent to Backtrace. Return undefined to skip an action, otherwise whatever (potentially modified) action returned will be sent to Backtrace.
+ */
+export const createBacktraceReduxMiddleware = (
+    client: BacktraceClient,
+    interceptAction: (action: Action) => Action | undefined = (action) => action,
```
perf2711, line 11: Can we make this optional?
```suggestion
    interceptAction?: (action: Action) => Action | undefined = (action) => action,
```

perf2711 (changes requested):

konraddysput (commented):
Can you please add a sample code?

packages/browser/package.json
adamcronin42: I'm torn, I feel like it may not need to be a `peerDependency`. For the actual sdk, all we really need are the types from `redux` which is included in `@reduxjs/toolkit`. I have `@reduxjs/toolkit` included as a `devDependency` to provide the necessary types and also so I can create a redux store, reducer, and actions in the tests. Thoughts?

> externalize it in Webpack
Would you mind elaborating?

packages/browser/src/redux/BacktraceReduxMiddleware.ts
adamcronin42: Since it has a default, isn't it already optional?

adamcronin42:
@konraddysput for sure! I'll add it to the browser example app

packages/browser/src/redux/BacktraceReduxMiddleware.ts
perf2711: oh, right, didn't see that one

packages/browser/package.json
perf2711: If we use only types from redux, then I would ensure that they are imported just as types:
```
import type { Middleware, Action } from 'redux';
```

Can you do this on every import from redux?

adamcronin42: Fixed. I made sure they are only imported as types. I believe those are the only two type imports from redux.

perf2711 (approved):

packages/browser/src/redux/BacktraceReduxMiddleware.ts
```diff
@@ -0,0 +1,32 @@
...
+) => {
+    if (!client) {
+        throw new Error('Must pass a BacktraceClient to the BacktraceReduxMiddleware.');
+    }
+    const middleware: Middleware = (store) => (next) => (action: Action) => {
```
konraddysput, line 17: The store variable is unused. If for any reason we have to keep it, can we add a linter ignore?

packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+const internalError = new Error('Test internal error');
+
+const testSlice = createSlice({
+    name: 'test',
+    initialState,
+    reducers: {
+        addToTestArray(state, action: PayloadAction<string>) {
+            state.testArr.push(action.payload);
+        },
+        toggleTestBool(state, action: PayloadAction<undefined>) {
```
konraddysput, line 35: The store variable is unused. If for any reason we have to keep it, can we add a linter ignore?

konraddysput (approved):

packages/browser/src/redux/BacktraceReduxMiddleware.ts
adamcronin42: Fixed

packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
adamcronin42: Fixed

PR #75, opened 2023-08-16
Fix possible circular reference in the annotation object

konraddysput:
# Why

Some annotations might contain circular references - for example, expresses's request/response object stores a reference to each other. Because of that, the JSON converter fails with a type error when we try to serialize the report. 

This diff fixes the issue with serialization and allows us to put in the JSON object "Circular reference" if we found it.


packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
```diff
@@ -44,13 +45,16 @@ export class MetricsSubmissionQueue<T extends MetricsEvent> implements MetricsQu
...
+                    {
+                        ...this._metricMetadata,
+                        [this._eventName]: events,
+                        metadata: {
+                            dropped_events: this._numberOfDroppedRequests,
+                        },
                     },
-                }),
+                    jsonEscaper(),
+                ),
```
perf2711, line 57: Maybe stringify once before the `for` loop?

perf2711 (approved):

packages/sdk-core/src/modules/metrics/MetricsSubmissionQueue.ts
konraddysput: good idea. Thx!

packages/sdk-core/src/common/jsonEscaper.ts
```diff
@@ -0,0 +1,28 @@
+export function jsonEscaper() {
+    const ancestors: unknown[] = [];
+    const keys: string[] = [];
+
+    // in TypeScript add "this: any" param to avoid compliation errors - as follows
+    //    return function (this: any, field: any, value: any) {
```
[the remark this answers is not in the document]
konraddysput: nope, we don't. I'm removing it. THanks

adamcronin42 (approved):

PR #77, opened 2023-08-16
Backtrace report unhandled error/crash overrides

konraddysput:
# Why

This diff extends unhandled error/crash report flow. In the past, the way how the unhandled rejection/error flow generated a not ideal classifier/message/annotations based on the error object (or conversion to the error object). 

With this change we can now:
*  override in the same way classifiers (and sets them when we need to), 
* timestamps

and because of this diff now:
* all unhandled promise rejections have classifier `UnhandledPromiseRejection` 
* the error object is not included if the user uses the reject API and returns a non-error object (for example reject('foo'))
*

### DIFF
--- a/examples/sdk/node/src/index.ts
+++ b/examples/sdk/node/src/index.ts
@@ -44,9 +44,10 @@ async function sendMessage(message: string, attributes: Record<string, number>) 
 }
 
 async function rejectPromise(message: string) {
-    return new Promise(() => {
+    return new Promise((_, rej) => {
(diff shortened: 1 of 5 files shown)


perf2711 (commented):
If we support multiple classifiers, why not add `error.name` to them when possible?

konraddysput:
> If we support multiple classifiers, why not add `error.name` to them when possible?

perf2711 (approved):

PR #78, opened 2023-08-16
Allow to use only one client via user-friendly API

konraddysput:
# Why

We can only have one crash-reporting solution available in the application created/spawned by us! In addition to that, if someone initialize the client once again, we might have a race condition between Backtrace Database instances. To avoid those problems the initialize method caches now the instance. With this change we know there is only one existing instance of the client. 

We still leave an option to create a backtrace client by the super user if there is a use case and benefit from doing that.

### DIFF
--- a/packages/react/src/BacktraceClient.ts
+++ b/packages/react/src/BacktraceClient.ts
@@ -5,7 +5,13 @@ export class BacktraceClient extends BrowserClient {
     public static builder(options: BacktraceConfiguration): BacktraceReactClientBuilder {
         return new BacktraceReactClientBuilder(options);
     }
-
+    /**
+     * Initializes the client. If the client already exists, the available instance
+     * will be returned and all other options will be ignored.
+     * @param options client configuration
+     * @param build builder
+     * @returns backtrace client
+     */
     public static initialize(options: BacktraceConfiguration, build?: (builder: BacktraceReactClientBuilder) => void) {
         if (this._instance) {
             return this._instance;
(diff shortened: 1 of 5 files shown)


packages/react/src/BacktraceClient.ts
```diff
@@ -1,24 +1,32 @@
...
-        if (!this._instance) {
-            throw new Error('BacktraceClient is uninitialized. Call "BacktraceClient.initialize" function first.');
-        }
+    /**
+     * Returns created BacktraceClient instance if the instance exists.
+     * Otherwise undefined.
+     */
+    public static get instance(): BacktraceClient | undefined {
         return this._instance;
     }
```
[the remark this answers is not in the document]
perf2711: When you would ever not have an instance available if you're using Backtrace? I think checking if instance exists is pointless and won't be a problem ever in runtime - only if the programmer makes an error and forgets to initialize Backtrace in time.

perf2711 (approved):

packages/react/src/BacktraceClient.ts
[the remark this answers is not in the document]
konraddysput: - when the user disables the error reporting integration and the app does not initialize them at all
- when you already have code like - if exists send it

avoid adding try/catch improves the user experience

packages/react/src/BacktraceClient.ts
```diff
@@ -1,24 +1,32 @@
-import { BacktraceConfiguration, BacktraceClient as BrowserClient } from '@backtrace/browser';
+import { BacktraceClient as BrowserClient, BacktraceConfiguration } from '@backtrace/browser';
 import { BacktraceReactClientBuilder } from './builder/BacktraceReactClientBuilder';
 
 export class BacktraceClient extends BrowserClient {
```
adamcronin42, line 4: I think the `private static _instance?: BacktraceClient;` accidentally got deleted

konraddysput: oh, no - it's available via browser backtrace client.

adamcronin42: Ohhh, that's right. My bad!

adamcronin42 (approved):

PR #81, opened 2023-08-17
Package export

konraddysput:
# Why

In the previous diff we prepared packages to contain files we want to have in the package (no src/utilities files). In this diff, I've fixed the issue with export/build process. What exactly?
- some exports were missing in the sdk-core/browser/react/node package
- the build process didn't build examples
- packages also required reference to packages already bundled - this solved theoretically the issue with missing type definition.


packages/browser/src/index.ts
```diff
@@ -1,8 +1,19 @@
...
+    BacktraceStackFrame,
+    BacktraceStackTraceConverter,
+    BacktraceStringAttachment,
+    BacktraceUint8ArrayAttachment,
+    BreadcrumbLogLevel,
+    BreadcrumbsEventSubscriber,
+    BreadcrumbsManager,
+    BreadcrumbType,
+    JavaScriptEngine,
+} from '@backtrace/sdk-core';
```
perf2711, line 15: maybe just export everything?
```suggestion
export * from '@backtrace/sdk-core';
```

perf2711 (approved):

packages/browser/src/index.ts
konraddysput: unfortunately there might be a naming issue - example: backtraceConfiguration

adamcronin42 (approved):

PR #83, opened 2023-08-18
Feature/redux example

adamcronin42:
Adding a redux example to our browser example app, a simple way to run the app (via serve), and some styling updates. To access the new redux example, click the 'Redux demo' button and you will see this screen:
image

### DIFF
--- a/examples/sdk/browser/src/index.ts
+++ b/examples/sdk/browser/src/index.ts
@@ -1,5 +1,6 @@
-import { BacktraceClient, BacktraceStringAttachment } from '@backtrace/browser';
+import { BacktraceClient, BacktraceStringAttachment, createBacktraceReduxMiddleware } from '@backtrace/browser';
 import { SUBMISSION_URL } from './consts';
+import { createSlice, configureStore } from '@reduxjs/toolkit';
 
 const client = BacktraceClient.initialize({
     url: SUBMISSION_URL,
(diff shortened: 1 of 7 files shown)


examples/sdk/browser/src/index.ts
```diff
@@ -14,6 +15,40 @@ const client = BacktraceClient.initialize({
...
+    reducers: {
+        setCount(state, action) {
+            state.count = action.payload;
+        },
+    },
+});
+
+const { setCount } = counterSlice.actions;
+
+const store = configureStore({
```
konraddysput, line 38: isn't this a react specific library? Do we need to add it to the package.json?

konraddysput (approved):

examples/sdk/browser/src/index.ts
adamcronin42: I'm not quite sure what you mean. Redux.js is UI/framework agnostic (but it does tend to be used the most with react). All of this redux functionality is included in `@reduxjs/toolkit`. Is that what you mean?

konraddysput: This is what I mean - we do not have @reduxjs/toolkit inside the package.json definition

adamcronin42: It's there. It's a dependency in the browser example package.json.

konraddysput (approved):

PR #84, opened 2023-08-22
Feature/browser readme

adamcronin42:
Here's my take on updating the browser readme. It's definitely not perfect but let me know what you think! Happy to make any changes. @ianrice07 I wasn't sure if you'd want to be on this or not so feel free to treat this as optional.

Notes:

- I removed references to database/databaseconfig as I don't think it is required for browser (let me know if that's incorrect)
- I reference the existing source maps Sauce doc but Sebastian's PR should update it soon and it should be the same url (I think)

### DIFF
--- a/packages/browser/README.md
+++ b/packages/browser/README.md
@@ -37,7 +37,7 @@ const options: BacktraceConfiguration = {
 }
 
 // Initialize the client
-const client = BacktraceClient.builder(options).build();
+const client = BacktraceClient.initialize(options);
 
 // Send an error
 client.send(new Error("Something broke!"));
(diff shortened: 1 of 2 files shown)


packages/browser/README.md
```diff
@@ -37,12 +41,33 @@ const options: BacktraceConfiguration = {
...
 
+## Sending an Error
+By default, Backtrace will send an error for Uncaught Exceptions and Unhandled Promise Rejections.
+
+### Manually Sending an Error
+There are several ways to send an error to Backtrace. For more details on the definition of ```client.send()``` see [Methods](#methods) below.
+
+```ts
+// send as a string
+await client.send('This is a string!');
```
konraddysput, line 58: Maybe instead of this we can add here info
"this is the message report"?

konraddysput (approved):

packages/browser/README.md
adamcronin42: My only concern is it might be confusing calling is a report when we have examples creating actual instances of BacktraceReport below. Thoughts?

PR #86, opened 2023-08-23
Adding react README.md

adamcronin42:
The only difference between this one and the browser readme is I updated all references of browser to react and added an Error Boundary section.


rick-bt (approved):
Great! But maybe move the error boundary section above source maps?

PR #89, opened 2023-08-24
Fallback to default log function if the util function is not available

konraddysput:
# Why

The breadcrumbs integration uses the util format method to format all console parameters. Util is the library available in the web browser or in node. However there is a chance the library won't be available in the sandbox env like in react-native. This diff adds a fallback function to format console logs when the util function is not available


adamcronin42 (approved):
I had never seen `private _formatter!:` syntax before. Pretty cool.

PR #94, opened 2023-08-25
Client dispose

konraddysput:
# Why

This diff adds dispose support to the client. 

Disposed client shouldn't be used or shouldn't send/generate any data.


adamcronin42 (commented):
What's the scenario where this would be called? Would we want to provide the user the ability to reinitialize? Currently, if the user calls initialize again, they will just get the disposed client. Is that what we want?

adamcronin42 (approved):

PR #95, opened 2023-08-25
Builder Api improvements

konraddysput:
# Why

We have a builder. The Builder class has limited options. We can't do too much with it, but with this change, we should override most of the options provided to the client. These changes are also needed in order to set up @backtrace/react-native client with a custom attributes system and without issues related to accessing document//window variables.

### DIFF
--- a/packages/browser/src/builder/BacktraceClientBuilder.ts
+++ b/packages/browser/src/builder/BacktraceClientBuilder.ts
@@ -32,14 +32,14 @@ export class BacktraceClientBuilder extends BacktraceCoreClientBuilder<Backtrace
             new UserIdentifierAttributeProvider(options),
             new ApplicationInformationAttributeProvider(options),
         ],
-        breadcrumbSubscribers: BreadcrumbsEventSubscriber[] = [
+        breadcrumbsSubscribers: BreadcrumbsEventSubscriber[] = [
             new WebRequestEventSubscriber(),
             new DocumentEventSubscriber(),
             new HistoryEventSubscriber(),
         ],
         sessionProvider: BacktraceSessionProvider = new BacktraceBrowserSessionProvider(),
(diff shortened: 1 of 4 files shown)


packages/browser/src/builder/BacktraceClientBuilder.ts
```diff
@@ -16,27 +23,33 @@ import { SpiderMonkeyStackTraceConverter } from '../converters/SpiderMonkeyStack
...
+    constructor(
+        protected readonly options: BacktraceConfiguration,
+        attributeProviders: BacktraceAttributeProvider[] = [
+            new UserAgentAttributeProvider(),
+            new WebsiteAttributeProvider(),
+            new WindowAttributeProvider(),
+            new UserIdentifierAttributeProvider(options),
+            new ApplicationInformationAttributeProvider(options),
+        ],
+        breadcrumbSubscribers: BreadcrumbsEventSubscriber[] = [
```
adamcronin42, line 35: May want to make this (and other references) plural.

adamcronin42 (approved):
Looks good! Left one comment.

packages/browser/src/builder/BacktraceClientBuilder.ts
konraddysput: it is plural, am I missing something?

adamcronin42: I was thinking breadcrumbSubscribers -> breadcrumb**s**Subscribers

adamcronin42 (approved):

packages/browser/src/builder/BacktraceClientBuilder.ts
konraddysput: Ah OK got it

PR #96, opened 2023-08-25
Fit 'n finish on node readme

rick-bt:
- Syntax updates
- Moved offline db sample up to initialization step
- Removed BacktraceDatabase detailed description in favor of a link to the object description


packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
```diff
@@ -2,7 +2,7 @@ export interface EnabledBacktraceDatabaseConfiguration {
     /**
```
rick-bt, line 2: Should these methods be EnableBacktraceDatabaseConfiguration and DisableBacktraceDatabaseConfiguration without the d?

konraddysput: I have mixed feelings about that. In the end the result should be the configuration for enabled database

PR #107, opened 2023-08-30
add examples packages as workspaces, remove package-lock.json from browser example

perf2711:


package.json
```diff
@@ -16,7 +16,8 @@
     },
     "workspaces": [
         "packages/*",
-        "tools/*"
```
konraddysput, line 19: this started breaking our packages - npm install packages based on directory order. This will also cause issues in the react-native demo, where react-native won't have access to packages needed to start the demo

I recommend to not do this now. We had this option but because of it we had a lot of troubles in the react/react-native project

konraddysput (changes requested):
Please let me know what you think about my comment

package.json
perf2711: All right, closing for now.

PR #110, opened 2023-08-31
add deferred user attributes

perf2711:
This PR adds a feature of deferred user attributes:
```ts
const client = BacktraceClient.initialize({
  userAttributes: () => { "a": 123 }
});

client.addAttributes(() => { "b": 456 });
```

Additionally, it changes the way the attributes are applied. Before this PR, the static attributes were always overridden by dynamic attributes, no matter the order. Now, the attributes are applied in the order as they are added:
* attribute providers
* user attributes
* report attributes

### DIFF
--- a/packages/sdk-core/src/BacktraceCoreClient.ts
+++ b/packages/sdk-core/src/BacktraceCoreClient.ts
@@ -1,5 +1,6 @@
 import {
     BacktraceAttachment,
+    BacktraceAttributeProvider,
     BacktraceConfiguration,
     BacktraceDatabaseRecord,
     BacktraceSessionProvider,
@@ -12,6 +13,7 @@ import { BacktraceReportSubmission } from './model/http/BacktraceReportSubmissio
 import { BacktraceReport } from './model/report/BacktraceReport';
 import { AttributeManager } from './modules/attribute/AttributeManager';
 import { ClientAttributeProvider } from './modules/attribute/ClientAttributeProvider';
+import { UserAttributeProvider } from './modules/attribute/UserAttributeProvider';
 import { BacktraceBreadcrumbs } from './modules/breadcrumbs';
 import { BreadcrumbsManager } from './modules/breadcrumbs/BreadcrumbsManager';
 import { V8StackTraceConverter } from './modules/converter/V8StackTraceConverter';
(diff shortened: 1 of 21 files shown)


packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -156,8 +165,14 @@ export abstract class BacktraceCoreClient {
      * Add attribute to Backtrace Client reports.
      * @param attributes key-value object with attributes.
      */
-    public addAttribute(attributes: Record<string, unknown>) {
-        this._attributeProvider.add(attributes);
+    public addAttribute(attributes: Record<string, unknown>): void;
+    public addAttribute(attributes: () => Record<string, unknown>): void;
+    public addAttribute(attributes: Record<string, unknown> | (() => Record<string, unknown>)) {
+        if (typeof attributes === 'function') {
```
konraddysput, line 171: What do you think about moving this detail to attribute manager? I think this should be the internal logic that belongs to the manager.

packages/sdk-core/src/modules/attribute/AttributeManager.ts
```diff
@@ -30,11 +26,11 @@ export class AttributeManager {
      */
     public async addProvider(attributeProvider: BacktraceAttributeProvider) {
```
konraddysput, line 27: actually maybe we can just keep add instead of addProvider and move the addprovider implementation to add ?

packages/sdk-core/src/modules/attribute/AttributeManager.ts
```diff
@@ -1,13 +1,9 @@
...
 import { ReportData } from '../../model/report/ReportData';
 import { BacktraceAttributeProvider } from './BacktraceAttributeProvider';
 import { ReportDataBuilder } from './ReportDataBuilder';
 
 export class AttributeManager {
-    public readonly attributes: Record<string, AttributeType> = {};
-    public readonly annotations: Record<string, unknown> = {};
-
-    private readonly _dynamicAttributeProviders: BacktraceAttributeProvider[] = [];
+    private readonly _attributeSources: Array<Record<string, unknown> | (() => Record<string, unknown>)> = [];
```
konraddysput, line 6: wouldn't it be better if we will keep all this data as AttributeProvider[]? the new type of the dynamic attribute is already an AttributeProvider with type dynamic. Keeping it in the previous model will be still useful and we will have one interface for everything . Right now we need to verify if something is a function or not

konraddysput (approved):

packages/sdk-core/src/modules/attribute/AttributeManager.ts
perf2711: I would keep them separate. `add` could be used to just add attributes or function resolving to attributes, and `addProvider` will just accept `BacktraceAttributeProvider`.

packages/sdk-core/src/modules/attribute/UserAttributeProvider.ts
```diff
@@ -0,0 +1,21 @@
+import { BacktraceAttributeProvider } from './BacktraceAttributeProvider';
+
+export class UserAttributeProvider implements BacktraceAttributeProvider {
+    constructor(private readonly _source: Record<string, unknown> | (() => Record<string, unknown>)) {}
+
+    public get type(): 'scoped' | 'dynamic' {
+        return typeof this._source === 'function' ? 'dynamic' : 'scoped';
```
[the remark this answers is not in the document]
perf2711: Changed.

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -156,8 +165,14 @@ export abstract class BacktraceCoreClient {
      * Add attribute to Backtrace Client reports.
      * @param attributes key-value object with attributes.
      */
-    public addAttribute(attributes: Record<string, unknown>) {
-        this._attributeProvider.add(attributes);
+    public addAttribute(attributes: Record<string, unknown>): void;
+    public addAttribute(attributes: () => Record<string, unknown>): void;
```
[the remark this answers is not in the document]
perf2711: Added a jsdoc and updated READMEs.

packages/sdk-core/src/modules/attribute/AttributeManager.ts
perf2711: Good idea. Changed.

konraddysput (approved):

adamcronin42 (approved):

PR #112, opened 2023-08-31
Detect unexpected message type and convert it to string

konraddysput:
# Why

We are all using javascript. Expect unexpected. In this situation, we discovered a possible error with an object error message. We need to handle this situation better. This diff making sure the message is always a string

### DIFF
--- a/packages/sdk-core/src/model/report/BacktraceReport.ts
+++ b/packages/sdk-core/src/model/report/BacktraceReport.ts
@@ -1,3 +1,4 @@
+import { jsonEscaper } from '../../common/jsonEscaper';
 import { TimeHelper } from '../../common/TimeHelper';
 import { BacktraceAttachment } from '../attachment';
 import { BacktraceStackFrame } from '../data/BacktraceStackTrace';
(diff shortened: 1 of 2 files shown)


perf2711 (approved):

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -109,4 +110,8 @@ export class BacktraceReport {
             this.classifiers.unshift(...options.classifiers);
         }
     }
+
+    private generateErrorMessage(data: unknown) {
+        return typeof data === 'object' ? JSON.stringify(data, jsonEscaper()) : data?.toString() ?? '';
```
adamcronin42, line 115: Should we maybe make the default something like 'Unknown error message'?

adamcronin42 (approved):
Nice! Left one comment

packages/sdk-core/src/model/report/BacktraceReport.ts
konraddysput: Can we leave it for now as it is and then adjust the flow when we recognize customers want to have something else? If someone puts undefined or null for any purpose - it would be better to show an empty message to show something went wrong

adamcronin42: Works for me!

PR #113, opened 2023-08-31
Inner report support

konraddysput:
# Why

this allows to send inner reports via the client


adamcronin42 (approved):

konraddysput:
The diff requires more clarification. Closing for now

PR #118, opened 2023-09-01
cli: add possibility of specifying source files for upload and add-sources

perf2711:
Currently `upload` and `add-sources` accept sourcemaps, which can be a bit misleading when using `process`, which accepts sources.

This PR allows for `upload` and `add-sources` to accept source files. CLI will try to resolve sourcemaps for given source files and continue as before:
```
> backtrace-js add-sources *.js
> backtrace-js process *.js
> backtrace-js upload *.js
> backtrace-js run *.js
```

### DIFF
--- a/tools/sourcemap-tools/src/SourceProcessor.ts
+++ b/tools/sourcemap-tools/src/SourceProcessor.ts
@@ -155,7 +155,7 @@ export class SourceProcessor {
     public getSourceMapPathFromSource(source: string, sourcePath: string) {
         const match = source.match(/^\/\/# sourceMappingURL=(.+)$/m);
         if (!match || !match[1]) {
-            return Err('Could not find source map for source.');
+            return Err('could not find source map for source.');
         }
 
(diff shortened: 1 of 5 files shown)


tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -155,7 +155,7 @@ export class SourceProcessor {
     public getSourceMapPathFromSource(source: string, sourcePath: string) {
         const match = source.match(/^\/\/# sourceMappingURL=(.+)$/m);
         if (!match || !match[1]) {
-            return Err('Could not find source map for source.');
+            return Err('could not find source map for source.');
```
konraddysput, line 158: Shouldn't it be upper case?

konraddysput (approved):

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: I'm trying to keep all errors lower case

adamcronin42 (approved):

PR #119, opened 2023-09-04
React native support

konraddysput:
# Why 

This diff shows the basic react-native support created based on existing features from the JavaScript monorepo


packages/react-native/android/src/main/java/backtrace/library/AttributeReader.java
```diff
@@ -0,0 +1,43 @@
...
+
+            while ((line = br.readLine()) != null) {
+                String[] entry = line.split(":", 2);
+                String key = entry[0].trim();
+                if(!attributeMapping.containsKey(key)){
+                    continue;
+                }
+                key = attributeMapping.get(key);
+                String value = entry[1].trim();
+                if(value.endsWith("kB")){
```
perf2711, line 31: Will that not affect attributes that may have a value ending with "kB"? Like "fookB"?

packages/react-native/android/src/main/java/backtrace/library/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.StringWriter;
+
+/**
+ * Handle unhandled Android exceptions from background threads.
+ */
+@ReactModule(name = backtraceio.library.BacktraceAndroidBackgroundUnhandledExceptionHandler.NAME)
+public class BacktraceAndroidBackgroundUnhandledExceptionHandler extends ReactContextBaseJavaModule implements Thread.UncaughtExceptionHandler  {
+    private final static transient String LOG_TAG = BacktraceAndroidBackgroundUnhandledExceptionHandler.class.getSimpleName();
+
+    private Thread.UncaughtExceptionHandler mRootHandler;
```
perf2711, line 24: Can we not do Hungarian notation?

packages/react-native/android/src/main/java/backtrace/library/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private Thread _lastCaughtBackgroundExceptionThread;
+    private Throwable _lastCaughtBackgroundException;
+
+    /**
+     * Check if data shouldn't be reported.
+     */
+    private volatile boolean shouldStop = false;
```
perf2711, line 32: Why some veriables begin with _ and some do not?

packages/react-native/android/src/main/java/backtrace/library/BacktraceReactNative.java
```diff
@@ -0,0 +1,104 @@
...
+
+        String[] keys = attributes.keySet().toArray(new String[0]);
+        String[] values = attributes.values().toArray(new String[0]);
+
+//        String[] attachments = new String[attachmentPaths.size()];
+//        for (attachmentPath:
+//             attachmentPaths.toArrayList()) {
+//            attachments[]
+//        }
+//        String[] attachments = attachmentPaths.toArrayList();
```
perf2711, line 67: is this required?

packages/react-native/android/src/main/java/backtrace/library/BacktraceDatabase.java
```diff
@@ -0,0 +1,12 @@
+package backtraceio.library;
+
+public class BacktraceDatabase {
```
perf2711, line 3: Can we rename it to something else? RIght now, I'm not sure if this is the same database that we have in the JS layer or not.

packages/react-native/ios/BacktraceApi.mm
```diff
@@ -0,0 +1,104 @@
+#import <Foundation/Foundation.h>
```
perf2711, line 1: Can we format this file? It's very unreadable.

packages/react-native/src/attributes/NativeAttributeProvider.ts
```diff
@@ -0,0 +1,19 @@
+import { type AttributeType, type BacktraceAttributeProvider } from '@backtrace-labs/sdk-core';
+import { NativeModules } from 'react-native';
+export class NativeAttributeProvider implements BacktraceAttributeProvider {
+    public get type(): 'scoped' | 'dynamic' {
+        return this._type;
+    }
```
perf2711, line 6: You can rename `_type` in constructor to `type` and change it's accessor to `public`. This getter will be redundant then.

packages/react-native/src/handlers/UnhandledExceptionHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+                        new BacktraceReport(
+                            rejection as Error,
+                            {
+                                'error.type': 'Unhandled exception',
+                                unhandledPromiseRejectionId: id,
+                            },
+                            [],
+                            {
+                                classifiers: ['UnhandledPromiseRejection'],
+                                skipFrames: rejection instanceof Error ? 0 : 1,
```
perf2711, line 45: On line 37 you cast to `Error`, but you check here whether `rejection` is `instanceof Error`. So it could be something else than error? Maybe we should employ additional checks on line 37?

Also applies to `rejectionTracking` mechanism.

packages/react-native/src/handlers/UnhandledExceptionHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+                    let stack: string | undefined;
+
+                    const stringValue = Object.prototype.toString.call(rejection);
+                    if (stringValue === '[object Error]') {
+                        message = Error.prototype.toString.call(rejection);
+                        stack = rejection.stack as string;
+                    } else {
+                        try {
+                            // eslint-disable-next-line @typescript-eslint/no-var-requires
+                            message = require('pretty-format')(rejection);
```
perf2711, line 84: Can we require it once?

For example:
```
function cachedPrettyFormat() {
  let prettyFormat: any | false;
  return function cachedPrettyFormat() {
    if (prettyFormat === false) {
      return undefined;
    }

    if (prettyFormat !== undefined) {
      return prettyFormat;
    }

    try {
      return (prettyFormat = require('pretty-format'));
    } catch {
      prettyFormat = false;
      return undefined;
    }    
  }
}

const getPrettyFormat = cachedPrettyFormat();

// then

const prettyFormat = getPrettyFormat();
if (prettyFormat) {
  message = prettyFormat(rejection);
} else {
  message = typeof rejection === 'string' ? rejection : JSON.stringify({ ...rejection });
}
```

packages/react-native/src/handlers/android/AndroidUnhandledExceptionHandler.ts
```diff
@@ -0,0 +1,42 @@
...
+    public captureManagedErrors(client: BacktraceClient) {
+        super.captureManagedErrors(client);
+        if (!this._unhandledExceptionHandler) {
+            return;
+        }
+        if (DebuggerHelper.isConnected()) {
+            return;
+        }
+
+        this._unhandledExceptionHandler.start(async (classifier: string, message: string, stackTrace: string) => {
```
perf2711, line 17: Can we skip `async/await`? Remember what it does to stack frames.

packages/react-native/src/handlers/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { AndroidUnhandledExceptionHandler } from './android/AndroidUnhandledExceptionHandler';
+import { type ExceptionHandler } from './ExceptionHandler';
+import { UnhandledExceptionHandler } from './UnhandledExceptionHandler';
+
+export function generateUnhandledExceptionHandler(): ExceptionHandler {
+    switch (Platform.OS) {
+        case 'android': {
+            return new AndroidUnhandledExceptionHandler();
+        }
+        default: {
```
perf2711, line 11: No iOS love? 😢

packages/react-native/android/src/main/java/backtrace/library/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
konraddysput: This is the style from the java code. I would prefer to stick to java code convention rather than using javascript code style here

packages/react-native/android/src/main/java/backtrace/library/BacktraceReactNative.java
konraddysput: Yes, we need to add a todo here.

packages/react-native/android/src/main/java/backtrace/library/BacktraceDatabase.java
konraddysput: Unfortunately no, without changes in our NDK library. Unfortunately We need to stick to class names available in the BacktraceAndroid. React always prefer to add a library prefix and NDK cannot detect exports without the library namespace

packages/react-native/src/handlers/index.ts
konraddysput: Every swift/objective-c error will cause an application crash. In Java managed errors might cause it, but crashpad won't capture them. When the managed environment detects an error we can still call javascript env and generate a report. On iOS the same thing is impossible.


There is also a change we extend the offer by adding more platforms later, so I believe its useful to have it here

packages/react-native/src/handlers/android/AndroidUnhandledExceptionHandler.ts
konraddysput: I forgot! Yes

packages/react-native/src/attributes/NativeAttributeProvider.ts
konraddysput: Good idea. Changed

packages/react-native/android/src/main/java/backtrace/library/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
konraddysput: After revisiting the code I adjusted it to your suggestions! Thank yoU!

packages/react-native/android/src/main/java/backtrace/library/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
konraddysput: After revisiting the code I adjusted it to your suggestions! Thank you!

packages/react-native/android/src/main/java/backtrace/library/AttributeReader.java
konraddysput: the values we're parsing always represent numbers - we have a list of attributes, we're collecting and the pattern is known.

packages/react-native/src/handlers/UnhandledExceptionHandler.ts
konraddysput: The code you see here is a copy-paste code from the react-native repository to just do literally all steps the same way they did to easily follow up what they did and what are we doing. Reference:

https://github.com/facebook/react-native/blob/a59b947a1e077d1c3f0d36926d374db5fe7d3291/packages/react-native/Libraries/promiseRejectionTrackingOptions.js#L38
Anyway I believe its useful to have this change so:
- I will add your change 
- I will add a reference to the code I'm talking about

packages/react-native/src/handlers/UnhandledExceptionHandler.ts
konraddysput: Adjusted the code. Thank you!

perf2711 (approved):

adamcronin42 (approved):

packages/react-native/package.json
```diff
@@ -0,0 +1,98 @@
...
+    },
+    "devDependencies": {
+        "@react-native-community/eslint-config": "^3.0.2",
+        "@types/react": "~17.0.21",
+        "@types/react-native": "0.70.0",
+        "pod-install": "^0.1.0",
+        "prettier": "^2.0.5",
+        "react": "18.2.0",
+        "react-native": "0.72.4",
+        "react-native-builder-bob": "^0.21.3",
```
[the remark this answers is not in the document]
konraddysput: metro is available in the react native demo. bob is a build system to build the library for all react native environments

PR #124, opened 2023-09-07
cli: add asset error behavior

perf2711:
This diff adds `--asset-error-behavior` which can control what happens when an asset fails:
* `exit` - process exits with code 1 as soon as an asset fails. This is the default behavior,
* `skip` - the asset is skipped silently,
* `error`, `warn`, `info`, `debug`, `trace` - the asset error is logged under this log level.

Note that if assets fail, the process can still exit with error code 1 - for example, if none assets succeed in upload, the upload will fail because no files were supplied.

### DIFF
--- a/tools/cli/src/helpers/errorBehavior.ts
+++ b/tools/cli/src/helpers/errorBehavior.ts
@@ -1,0 +1,50 @@
+import { Err, Ok, Result, ResultErr } from '@backtrace-labs/sourcemap-tools';
+
+export const ErrorBehaviors = {
+    exit: 'exit',
+    error: 'error',
+    warn: 'warn',
+    info: 'info',
+    debug: 'debug',
+    trace: 'trace',
+    skip: 'skip',
+} as const;
+
+export type ErrorBehavior = keyof typeof ErrorBehaviors;
+
+export function GetErrorBehavior(type: string): Result<ErrorBehavior, string> {
+    const valid = Object.keys(ErrorBehaviors);
+    if (valid.includes(type)) {
+        return Ok(type as ErrorBehavior);
+    }
+
+    return Err(`invalid error behavior "${type}", expected one of: ${valid.join(', ')}`);
+}
+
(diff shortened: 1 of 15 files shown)


tools/cli/src/helpers/errorBehavior.ts
```diff
@@ -0,0 +1,50 @@
...
+    warn: 'warn',
+    info: 'info',
+    debug: 'debug',
+    trace: 'trace',
+    skip: 'skip',
+} as const;
+
+export type ErrorBehavior = keyof typeof ErrorBehaviors;
+
+export function GetErrorBehavior(type: string): Result<ErrorBehavior, string> {
```
konraddysput, line 15: functions should start with lower case.

tools/cli/src/helpers/errorBehavior.ts
```diff
@@ -0,0 +1,50 @@
...
+    const valid = Object.keys(ErrorBehaviors);
+    if (valid.includes(type)) {
+        return Ok(type as ErrorBehavior);
+    }
+
+    return Err(`invalid error behavior "${type}", expected one of: ${valid.join(', ')}`);
+}
+
+export interface FailedElement<E> {
```
konraddysput, line 24: isn't FailedElement too generic? Wouldn't it be better if we rename it to FailedErrorBehavior?

konraddysput (approved):
Before we land - can we confirm if it works with the run command if it works, what is the expected run json object?

tools/cli/src/helpers/errorBehavior.ts
perf2711: Renamed to `BehaviorSkippedElement`.

tools/cli/src/helpers/errorBehavior.ts
perf2711: Changed.

adamcronin42 (approved):

PR #130, opened 2023-09-11
cli: optimization of `run` and simplification of code

perf2711:
This PR changes some stuff in sourcemap-tools and CLI:
* remove `AsyncResult` and use `pipe` / `flow` instead
* add `--include` and `--exclude` functionality
* `run` now calls functions with loaded sources and sourcemaps instead of passing paths
* `run` now writes files instead of each command

I apologize for the size of this PR - changes there and there were breaking other things.


tools/cli/src/helpers/common.ts
```diff
@@ -1,42 +1,114 @@
...
+                ),
+            ),
+        );
+    };
+}
+
+export function writeAsset<T extends AssetWithContent<unknown>>(asset: T) {
+    return pipe(
+        asset.content,
+        (content) => (typeof content === 'object' ? JSON.stringify(content) : String(content)),
```
konraddysput, line 60: I think you should use toString instead of String. 
```suggestion
        (content) => (typeof content === 'object' ? JSON.stringify(content) : content?.toString() ?? "",
```

tools/cli/src/options/loadOptions.ts
```diff
@@ -13,33 +13,26 @@ export function loadOptionsForCommand(path?: string) {
...
-        }
-
-        return AsyncResult.equip(readResult)
-            .then(parseJSONC<CliOptions>)
-            .then((opts) => (readOptions = opts))
-            .then(joinOptions(key, defaults)).inner;
+        return pipe(
+            path,
+            (path) => (!readOptions ? loadOptions(path) : Ok(readOptions)),
+            R.map((data) => (data ? (readOptions = data) : data)),
```
konraddysput, line 19: how it works - you're assigning something when data is defined, but you're not returning it and in the next method you're trying to use it. Are you sure it works correctly? I think the result will be undefined here.

tools/cli/src/sourcemaps/add-sources.ts
```diff
@@ -117,92 +139,68 @@ export async function addSourcesToSourcemaps({ opts, logger, getHelpMessage }: C
...
+            ),
+            R.mapErr((error) => `${asset.name}: ${error}`),
+            handleFailedAsset(logAssetBehaviorError(asset)),
+        );
+
+    const includePaths = normalizePaths(opts.include);
+    const excludePaths = normalizePaths(opts.exclude);
+    const { isIncluded, isExcluded } = await buildIncludeExclude(includePaths, excludePaths, logTrace);
+
+    return pipe(
```
konraddysput, line 168: what happened to the rest commands available in the previous version? Why we don't need them anymore?

konraddysput: ah ok, we have file per command now (feel free to correct me if I'm wrong)

konraddysput (approved):

tools/cli/src/helpers/common.ts
```diff
@@ -1,42 +1,114 @@
...
+        writeFile(asset.path),
+        R.map(() => asset),
+    );
+}
+
+export function writeSourceAndSourceMap<T extends SourceAndSourceMap>(asset: T) {
+    return pipe(
+        asset,
+        () => pipe(asset.source.content, writeFile(asset.source.path)),
+        R.map(() => pipe(JSON.stringify(asset.sourceMap.content), writeFile(asset.sourceMap.path))),
```
konraddysput, line 70: shouldn't you use parseJson here instead?

konraddysput (approved):

tools/cli/src/sourcemaps/add-sources.ts
perf2711: Umm, we always had file per command? `add-sources`, `process`, `upload`, and `run` are available.

tools/cli/src/helpers/common.ts
perf2711: `parseJson` deserializes JSON, here I serialize it.

tools/cli/src/options/loadOptions.ts
perf2711: `(readOptions = data)` returns `data`:
image

tools/cli/src/helpers/common.ts
perf2711: Typescript is erroring here that `.toString()` is not defined.

adamcronin42 (approved):

PR #134, opened 2023-09-14
Single client instance

konraddysput:
# Why

Right now each client implements how to deal with the Backtrace "instance". The code is redundant. In addition to that storing the instance in a library "instance" instead of globally for every other instance might create some weird issues - for example: The error boundary won't find an instance. 

This diff fixes the behavior and forces all clients to use one single instance.

### DIFF
--- a/packages/browser/src/BacktraceClient.ts
+++ b/packages/browser/src/BacktraceClient.ts
@@ -18,7 +18,6 @@ import { BacktraceClientBuilder } from './builder/BacktraceClientBuilder';
 export class BacktraceClient extends BacktraceCoreClient {
     private readonly _disposeController: AbortController = new AbortController();
 
-    protected static _instance?: BacktraceClient;
     constructor(
         options: BacktraceConfiguration,
         requestHandler: BacktraceRequestHandler,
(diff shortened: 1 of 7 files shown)


packages/react/src/BacktraceClient.ts
```diff
@@ -43,21 +41,24 @@ export class BacktraceClient extends BrowserClient {
...
     }
 
     /**
      * Returns created BacktraceClient instance if the instance exists.
      * Otherwise undefined.
      */
     public static get instance(): BacktraceClient | undefined {
-        return this._instance;
+        return this._instance as BacktraceClient;
     }
```
perf2711, line 63: You should be able to move this to `BacktraceCoreClient` and mark it to return `this | undefined`.

konraddysput: goood idea.

konraddysput: I checked your idea and there is a problem with the instance variable type - we use it to make sure client is a singleton type. In this situation, we cannot type it to `this` 

image

perf2711 (approved):

packages/browser/src/BacktraceClient.ts
```diff
@@ -59,21 +58,21 @@ export class BacktraceClient extends BacktraceCoreClient {
         options: BacktraceConfiguration,
         build?: (builder: BacktraceClientBuilder) => void,
     ): BacktraceClient {
-        if (this._instance) {
-            return this._instance;
+        if (this.instance) {
```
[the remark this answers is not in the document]
konraddysput: No. Getter already cover casting the instance to the right type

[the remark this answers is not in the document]
adamcronin42: Ohhhh. Forgot about the getter. Thanks!

adamcronin42 (approved):

PR #135, opened 2023-09-14
Allow to modify error boundary by other clients

konraddysput:
# Why

Error boundary was designed for react use cases. However, the same error boundary should be used by the react-native client and react-native client needs to override/expose some information from it.

This diff changes:
* default name from "generic" to "backtrace" specific. Other people now instead of importing prop will import something that is not as generic as it was
* allow to use any backtrace client we want - needed if the instance in the react client is not available, but is in the react-native.

### DIFF
--- a/packages/react/src/ErrorBoundary.tsx
+++ b/packages/react/src/ErrorBoundary.tsx
@@ -4,30 +4,29 @@ import { BacktraceClient } from './BacktraceClient';
 
 type RenderFallback = (error: Error) => ReactElement;
 
-export interface Props {
+export interface BacktraceErrorBoundaryProps {
     children: ReactNode;
     fallback?: ReactElement | RenderFallback;
     name?: string;
 }
 
-export interface State {
+export interface BacktraceErrorBoundaryState {
     error?: Error;
 }
 
-export class ErrorBoundary extends Component<Props, State> {
-    private _client: BacktraceClient;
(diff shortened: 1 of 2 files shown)


packages/react/src/ErrorBoundary.tsx
```diff
@@ -4,30 +4,29 @@ import { BacktraceClient } from './BacktraceClient';
...
 }
 
-export class ErrorBoundary extends Component<Props, State> {
-    private _client: BacktraceClient;
+export class ErrorBoundary extends Component<BacktraceErrorBoundaryProps, BacktraceErrorBoundaryState> {
     private COMPONENT_THREAD_NAME = 'component-stack';
-    constructor(props: Props) {
+    constructor(
+        props: BacktraceErrorBoundaryProps,
+        private readonly _client: BacktraceClient = BacktraceClient.instance as BacktraceClient,
```
perf2711, line 21: I think you should not add anything outside `props` object when creating a React component.

perf2711 (changes requested):

packages/react/src/ErrorBoundary.tsx
konraddysput: the constructor with the client parameter will be used by error boundaries that inherit from the react error boundary. The client parameter doesn't matter in error boundary creation, but it matters in calling base class constructor from react-native

perf2711 (approved):

PR #136, opened 2023-09-14
Native Submission URLs

konraddysput:
# Why

Most of the time our libraries need to send json reports - this is why submission URL class has helpers around a json format. However other our libraries like react-native or electron will need to use different URLs to submit different types of data - like minidumps or plcrashreporter crashes. 

This diff extends the submission URL class and allow to transform the URL to the plcrashreporter or minidump format

### DIFF
--- a/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
+++ b/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
@@ -31,6 +31,22 @@ export class SubmissionUrlInformation {
     }
 
     /**
+     * Converts full submission JSON URL to PlCrashReporter submission URL
+     * @param submissionUrl Backtrace Submission URL
+     */
+    public static toPlCrashReporterSubmissionUrl(submissionUrl: string): string {
+        return this.changeSubmissionFormat(submissionUrl, 'minidump');
+    }
+
+    /**
+     * Converts full submission JSON URL to minidump submission URL
+     * @param submissionUrl Backtrace Submission URL
+     */
+    public static toMinidumpSubmissionUrl(submissionUrl: string): string {
+        return this.changeSubmissionFormat(submissionUrl, 'minidump');
+    }
+
+    /**
      * Find the universe based on the submission URL
      * @param submissionUrl submission URL
      * @returns universe name
@@ -70,5 +86,16 @@ export class SubmissionUrlInformation {
         const url = new URL(submissionUrl);
 
         return url.searchParams.get('token');
+    }
+
+    public static changeSubmissionFormat(submissionUrl: string, desiredFormat: 'json' | 'plcrash' | 'minidump') {
+        const submitIndex = submissionUrl.indexOf(this.SUBMIT_PREFIX);
+        const url = new URL(submissionUrl);
+        if (submitIndex !== -1) {
+            url.pathname = url.pathname.replace('/json', `/${desiredFormat}`);
+        } else {
+            url.searchParams.set('format', desiredFormat);
+        }
+        return url.href;
     }
 }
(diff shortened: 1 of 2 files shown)


packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -30,6 +30,22 @@ export class SubmissionUrlInformation {
         return result.href;
     }
 
+    /**
+     * Converts full submission JSON URL to PlCrashReporter submission URL
+     * @param submissionUrl Backtrace Submission URL
+     */
+    public static toPlCrashReporterSubmissionUrl(submissionUrl: string): string {
+        return this.changeSubmissionFormat(submissionUrl, 'minidump');
+    }
```
perf2711, line 39: it's a little bit weird to include this function here. It's highly vendor specific, and will be used only in React Native. As it is identical to `toMinidumpSubmissionUrl`, why not just keep that one?

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -71,4 +87,15 @@ export class SubmissionUrlInformation {
 
         return url.searchParams.get('token');
     }
+
+    public static changeSubmissionFormat(submissionUrl: string, desiredFormat: 'json' | 'plcrash' | 'minidump') {
+        const submitIndex = submissionUrl.indexOf(this.SUBMIT_PREFIX);
+        const url = new URL(submissionUrl);
+        if (submitIndex !== -1) {
+            url.pathname = url.pathname.replace('/json', `/${desiredFormat}`);
```
perf2711, line 95: How would you know if the submission URL had `/json` to begin with?

konraddysput: if it doesn't then we're dealing with the invalid URL and we cannot do anything with it

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
konraddysput: your suggestion is to keep only `changeSubmissionFormat` function are remove plcrashreporter and minidump functions?

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
perf2711: but this is not true. If you replace `/json` with `/minidump`, is this now an invalid URL? Because if I pass it, and ask the function to replace this back to `json`, it will not work.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
perf2711: Either that or keep `toMinidumpSubmissionUrl`.

konraddysput: ah ok - I see the issue. this is wrong - plcrashreporter submission URL should be plrash instead of minidump. Modified it to the desired format

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
konraddysput: Modified the algorithm to use the last part of the URL instead

perf2711: What if someone passes an URL with trailing slash?
https://submit.backtrace.io/name/000000000000a1eb7ae344f6e002de2e20c81fbdedf6991c2f3bb45b11111111/json/

I guess the implementation will break then.

konraddysput: For this specific use case I've changed it to pathname split

PR #140, opened 2023-09-15
Feature/attachment management

konraddysput:
# Why

Each Backtrace client might have a use case where we need to do something when an attachment is being added to the client. For example - on the react-native side, the attachment should be passed to the native client. The previous model where we exposed attachments via attachments array, didn't fit well this criteria. This diff adds the `addAttachment` method to the `BacktraceCoreClient` and attachments now can only be added via this method. Other clients can override it to be notified when an attachment was added by the client and do w/e they want

### DIFF
--- a/packages/sdk-core/src/BacktraceCoreClient.ts
+++ b/packages/sdk-core/src/BacktraceCoreClient.ts
@@ -79,18 +79,22 @@ export abstract class BacktraceCoreClient {
     }
 
     /**
+     * Client cached attachments
+     */
+    public get attachments(): BacktraceAttachment[] {
+        return this._attachments;
+    }
+
+    /**
      * Report database used by the client
      */
     public get database(): BacktraceDatabase | undefined {
         return this._database;
     }
 
-    /**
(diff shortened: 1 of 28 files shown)


packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -79,18 +79,23 @@ export abstract class BacktraceCoreClient {
     }
 
     /**
-     * Report database used by the client
+     * Client cached attachments
      */
-    public get database(): BacktraceDatabase | undefined {
-        return this._database;
+    public get attachments(): BacktraceAttachment[] {
```
perf2711, line 84: Make the return type `readonly BacktraceAttachment[]`.

perf2711 (approved):

packages/sdk-core/src/BacktraceCoreClient.ts
konraddysput: Changed to readonly

adamcronin42 (approved):

PR #147, opened 2023-09-22
sdk-core, node: add file system abstractions

perf2711:
This PR moves file system abstraction to sdk-core, allowing other clients to use this abstraction to achieve, for example, database support, without reimplementing the business logic.

Database file support has also been moved to sdk-core, which works with `FileSystem` interface now.

Jira task: BT-585


packages/node/src/BacktraceClient.ts
```diff
@@ -219,7 +221,7 @@ export class BacktraceClient extends BacktraceCoreClient {
     }
 
     private async loadNodeCrashes() {
-        if (!this.options.database?.captureNativeCrashes) {
+        if (!this.fileSystem || !this.options.database?.captureNativeCrashes) {
```
konraddysput, line 224: this is out of scope for this review, however I have a question:

- client is a programming interface for developer to use Backtrace. For managing database we have a separated class which is BacktraceDatabase. Shouldn't this code be a part of the database somehow?

packages/sdk-core/src/modules/storage/FileSystem.ts
```diff
@@ -0,0 +1,23 @@
+import { BacktraceAttachment } from '../../model/attachment';
+
+export interface FileSystem {
+    readDir(dir: string): Promise<string[]>;
```
konraddysput, line 4: quick idea:

based on the code we had some assumptions or we needed to pass some options - like for example:
- we expect to read the directory and return all files with extensions 
- we also expect to return files in specific string format - utf8

Shouldn't we note those assumptions in the interface to avoid potential surprises later? 
Previous code:

```
  const databaseFiles = fs.readdirSync(databasePath, {
            encoding: 'utf8',
            withFileTypes: true,
        });
```

packages/node/src/storage/NodeFileSystem.ts
```diff
@@ -0,0 +1,62 @@
...
+import fs from 'fs';
+import { BacktraceFileAttachment } from '../attachment';
+
+export class NodeFileSystem implements FileSystem {
+    public readDir(dir: string): Promise<string[]> {
+        return fs.promises.readdir(dir);
+    }
+
+    public readDirSync(dir: string): string[] {
+        return fs.readdirSync(dir);
```
konraddysput, line 11: previously you were using the following options to read the dir:

```
  const databaseFiles = fs.readdirSync(databasePath, {
            encoding: 'utf8',
            withFileTypes: true,
        });
```

Are you sure it works correctly without them?

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -71,11 +77,15 @@ export abstract class BacktraceCoreClient {
     }
 
     public get metrics(): BacktraceMetrics | undefined {
-        return this._metrics;
+        return this._modules.get(BacktraceMetrics);
```
konraddysput, line 80: did you merge your code from dev to this branch? You already landed this change

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -131,25 +143,27 @@ export abstract class BacktraceCoreClient {
...
             this._sdkOptions,
             stackTraceConverter,
             this.attributeManager,
             new DebugIdProvider(stackTraceConverter, this._setup.debugIdMapProvider),
         );
 
-        this._attachments = this.options.attachments ?? [];
+        if (this.options?.database?.enable === true) {
+            const provider =
+                this._setup.databaseStorageProvider ??
```
konraddysput, line 156: You can drop databaseStorageProvider support and use only a new file system. The core client interface is not exposed outside the package and also to construct it you need to have package internals not exposed anywhere else.

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -158,19 +172,38 @@ export abstract class BacktraceCoreClient {
...
+            this.attributeManager.addProvider(breadcrumbsManager);
+            this._modules.set(BreadcrumbsManager, breadcrumbsManager);
         }
 
-        this.initialize();
         this._enabled = true;
     }
 
+    public initialize() {
+        if (this.fileSystem && this.options.database?.createDatabaseDirectory) {
```
konraddysput, line 189: I think this should be a part of the BacktraceDatabase implementation.

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
```diff
@@ -24,19 +26,25 @@ export class BacktraceDatabaseFileRecord implements BacktraceDatabaseRecord {
     public static fromRecord(record: BacktraceDatabaseRecord) {
         return new BacktraceDatabaseFileRecord(
             record,
-            record.attachments.filter((n) => n instanceof BacktraceFileAttachment) as BacktraceFileAttachment[],
+            record.attachments.filter(BacktraceDatabaseFileRecord.isFileAttachment),
```
konraddysput, line 29: Isn't it cleaner we expect to pass File Attachment here instead of doing this inside the class? Now if someone will use it, it would be surprising for someone attachment are not there or are filtered out by the class implementation. WDYT?

konraddysput (changes requested):
Can we please merge changes from the dev/your previous branch?

packages/sdk-core/src/modules/database/BacktraceDatabaseFileStorageProvider.ts
```diff
@@ -76,48 +65,44 @@ export class BacktraceDatabaseFileStorageProvider implements BacktraceDatabaseSt
...
+            this._fileSystem.unlinkSync(recordPath);
             return true;
         } catch {
             return false;
         }
     }
 
     private getRecordPath(id: string): string {
-        return path.join(this._path, `${id}${this.RECORD_SUFFIX}`);
+        return this._path + '/' + `${id}${this.RECORD_SUFFIX}`;
```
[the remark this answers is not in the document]
perf2711: `path` is unavailable in `sdk-core`

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
```diff
@@ -24,19 +26,25 @@ export class BacktraceDatabaseFileRecord implements BacktraceDatabaseRecord {
...
 
-    public static fromJson(json: string): BacktraceDatabaseFileRecord | undefined {
+    public static fromJson(json: string, fileSystem: FileSystem): BacktraceDatabaseFileRecord | undefined {
         try {
             const record = JSON.parse(json) as BacktraceDatabaseFileRecord;
             const attachments = record.attachments
-                ? record.attachments.map((n) => new BacktraceFileAttachment(n.filePath))
+                ? record.attachments
+                      .filter(BacktraceDatabaseFileRecord.isFileAttachment)
+                      .map((n) => fileSystem.createAttachment(n.filePath, n.name))
```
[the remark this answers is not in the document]
perf2711: we do not have `BacktraceFileAttachment` class here, it is available in Node

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
```diff
@@ -24,19 +26,25 @@ export class BacktraceDatabaseFileRecord implements BacktraceDatabaseRecord {
...
                 : [];
             return new BacktraceDatabaseFileRecord(record, attachments);
         } catch {
             return undefined;
         }
     }
+
+    private static isFileAttachment(attachment: BacktraceAttachment): attachment is BacktraceFileAttachment {
```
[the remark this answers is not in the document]
perf2711: I can pass it more safely to `filter` as a function and not worry about it accessing `this`.

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: due to merging previous stuff, I had to rebase.

packages/sdk-core/src/modules/storage/FileSystem.ts
perf2711: So `withFileTypes` returns not extensions, but objects instead that have `isFile` or `isDir` methods. I opted for skipping that as we will have no directories being created with the names that we search for.

Also, `utf8` is default:
image

packages/node/src/storage/NodeFileSystem.ts
perf2711: yes

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: I'm not sure what you mean? Use `fileSystem` directly in `BacktraceDatabase`?

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
perf2711: not sure what you mean by this. I don't have `BacktraceFileAttachment` implementation here. But every implementatio of the interface will return `true` from this filter.

packages/node/src/BacktraceClient.ts
```diff
@@ -36,7 +36,7 @@ export class BacktraceClient extends BacktraceCoreClient {
             breadcrumbsSetup: {
                 subscribers: breadcrumbsEventSubscribers,
             },
-            databaseStorageProvider: BacktraceDatabaseFileStorageProvider.createIfValid(options.database),
+            fileSystem: new NodeFileSystem(),
```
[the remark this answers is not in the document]
perf2711: Done

perf2711:
> Can we please merge changes from the dev/your previous branch?

Merged.

packages/node/src/BacktraceClient.ts
perf2711: I think not, because depending all stuff on BacktraceDatabase that can be dependent just on FileSystem is wrong. It's completely unnecessary.

It makes sense for me that FileSystem is the dependency of BacktraceDatabase and other stuff that may use it, and not have the DB inbetween.
image

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: This is really troublesome. If we didn't rely on a database directory for everything, but just use something like "storageOptions", this would make perfect sense.

Moving it to database will cause undeterministic behavior, when some module depending on DB's directory will be initialized before DB.

packages/sdk-core/src/BacktraceCoreClient.ts
konraddysput: I think you're trying to use databaseStorageProvider - I think you can remove it and use only the new file system variable

packages/node/src/BacktraceClient.ts
konraddysput: but here you have something that clearly depends on the database. You literally need to check if the database is initialized and do something only when it is.

packages/node/src/storage/NodeFileSystem.ts
konraddysput: How so - we had fun use cases with our customers in India/asia having issues with default encoding. I think using utf8 is necessary. is this a default option?

packages/node/src/BacktraceClient.ts
perf2711: that's a business decision, not a code decision.

packages/node/src/storage/NodeFileSystem.ts
perf2711: `utf-8` is the default encoding in `readdir`/`readdirSync`

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: so we will not be planning to override providers from outside, that's what you mean?

konraddysput: is there any reason to do that if we have file system? I think, no

perf2711: Ok, removed

konraddysput (approved):

PR #149, opened 2023-09-26
cli: various fixes

perf2711:
This PR fixes some errors with CLI:
* bring back `subdomain` and `token` options
* display process and add-sources messages in run only if enabled
* fix loading config for subcommands
* fix dry-run and pass-with-no-files for upload
* update README

### DIFF
--- a/tools/cli/src/sourcemaps/upload.ts
+++ b/tools/cli/src/sourcemaps/upload.ts
@@ -97,7 +97,7 @@ export const uploadCmd = new Command<UploadOptions>({
     .option({
         name: 'subdomain',
         type: String,
-        description: 'Subdomain to upload to.',
+        description: 'Subdomain to upload to. Do not use on on-premise environments.',
         alias: 's',
     })
     .option({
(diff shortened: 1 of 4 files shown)


tools/cli/src/sourcemaps/upload.ts
```diff
@@ -346,11 +360,23 @@ function pipeAssets(assets: AssetWithContent<RawSourceMapWithDebugId>[], include
...
     if (opts.url) {
         return validateUrl(opts.url);
     }
 
+    if (opts.subdomain) {
+        if (!opts.token) {
+            return Err('token is required with subdomain');
+        }
+
+        return Ok(`https://submit.backtrace.io/${opts.subdomain}/${opts.token}/sourcemap`);
```
konraddysput, line 377: I think its worth to note the URL won't work on the on prem environment. What do you think about putting this note in notes?

konraddysput (approved):

tools/cli/src/sourcemaps/upload.ts
perf2711: Done

adamcronin42 (approved):

PR #150, opened 2023-09-26
node: breadcrumbs from previous session in database

perf2711:
This PR adds the functionality of storing breadcrumbs from previous session in database.

The way it works is with a new `SessionFiles` class, which manages all previous sessions. The sessions are cleared after initialization, however, a module may lock the session before to delay the clearing. After the module unlocks the session, it is cleared automatically. This helps to remove sessions that are not used by pending reports, and not to remove those that are used.

By default, the previous sessions are limited to 1. Meaning, that if more than one previous session is available, only the most recent one will be able to be locked.

The breadcrumbs are stored on disk as each line representing a JSON breadcrumb. It is not a JSON array. Two files are used to preserve breadcrumbs. After reaching maximum breadcrumbs in one file, the main file will be moved to a fallback file, and new breadcrumbs will be written to the main file. Both files will be uploaded to a crash.

Jira task: BT-585

### DIFF
--- a/packages/node/src/BacktraceClient.ts
+++ b/packages/node/src/BacktraceClient.ts
@@ -4,6 +4,7 @@ import {
     BacktraceReport,
     BacktraceRequestHandler,
     BreadcrumbsEventSubscriber,
+    BreadcrumbsManager,
     BacktraceConfiguration as CoreConfiguration,
     DebugIdContainer,
     FileSystem,
@@ -12,6 +13,7 @@ import {
 import path from 'path';
 import { BacktraceConfiguration } from './BacktraceConfiguration';
 import { AGENT } from './agentDefinition';
+import { FileBreadcrumbsStorage } from './breadcrumbs/FileBreadcrumbsStorage';
 import { BacktraceClientBuilder } from './builder/BacktraceClientBuilder';
 import { NodeOptionReader } from './common/NodeOptionReader';
 import { NodeDiagnosticReportConverter } from './converter/NodeDiagnosticReportConverter';
@@ -37,6 +39,13 @@ export class BacktraceClient extends BacktraceCoreClient {
             },
             fileSystem,
         });
+
+        const breadcrumbsManager = this.modules.get(BreadcrumbsManager);
+        if (breadcrumbsManager && this.sessionFiles) {
+            breadcrumbsManager.setStorage(
+                FileBreadcrumbsStorage.create(this.sessionFiles, options.breadcrumbs?.maximumBreadcrumbs ?? 100),
+            );
+        }
     }
 
     public initialize(): void {
--- a/packages/node/src/BacktraceClient.ts
+++ b/packages/node/src/BacktraceClient.ts
@@ -50,8 +50,7 @@ export class BacktraceClient extends BacktraceCoreClient {
     }
 
     public initialize(): void {
-        const previousSession = this.sessionFiles?.getPreviousSession();
-        const lockId = previousSession?.lock();
+        const lockId = this.sessionFiles?.lockPreviousSessions();
 
         try {
             super.initialize();
@@ -62,11 +61,11 @@ export class BacktraceClient extends BacktraceCoreClient {
 
             this.captureNodeCrashes();
         } catch (err) {
-            lockId && previousSession?.unlock(lockId);
+            lockId && this.sessionFiles?.unlockPreviousSessions(lockId);
             throw err;
         }
 
-        this.loadNodeCrashes(previousSession).finally(() => lockId && previousSession?.unlock(lockId));
+        this.loadNodeCrashes().finally(() => lockId && this.sessionFiles?.unlockPreviousSessions(lockId));
     }
 
     public static builder(options: BacktraceConfiguration): BacktraceClientBuilder {
@@ -238,7 +237,7 @@ export class BacktraceClient extends BacktraceCoreClient {
         }
     }
 
-    private async loadNodeCrashes(sessionFiles?: SessionFiles) {
+    private async loadNodeCrashes() {
         if (!this.database || !this.fileSystem || !this.options.database?.captureNativeCrashes) {
             return;
         }
@@ -266,17 +265,40 @@ export class BacktraceClient extends BacktraceCoreClient {
             return;
         }
 
+        const reports: [path: string, report: BacktraceReport, sessionFiles?: SessionFiles][] = [];
         for (const recordName of recordNames) {
             const recordPath = path.join(databasePath, recordName);
             try {
                 const recordJson = await this.fileSystem.readFile(recordPath);
                 const report = converter.convert(JSON.parse(recordJson));
+                reports.push([recordPath, report]);
+            } catch {
+                // Do nothing, skip the report
+            }
+        }
 
-                if (sessionFiles) {
-                    const breadcrumbsStorage = FileBreadcrumbsStorage.createFromSession(sessionFiles);
+        // Sort reports by timestamp descending
+        reports.sort((a, b) => b[1].timestamp - a[1].timestamp);
+
+        // Map reports to sessions
+        // When the sessions are sorted by timestamp, we can assume that each previous session maps to the next report
+        let currentSession = this.sessionFiles?.getPreviousSession();
+        for (const tuple of reports) {
+            tuple[2] = currentSession;
+            currentSession = currentSession?.getPreviousSession();
+        }
+
+        for (const [recordPath, report, session] of reports) {
+            try {
+                if (session) {
+                    const breadcrumbsStorage = FileBreadcrumbsStorage.createFromSession(session);
                     if (breadcrumbsStorage) {
                         report.attachments.push(...breadcrumbsStorage.getAttachments());
                     }
+
+                    report.attributes['application.session'] = session.sessionId;
+                } else {
+                    report.attributes['application.session'] = null;
                 }
 
                 const data = this.generateSubmissionData(report);
--- a/packages/node/src/BacktraceClient.ts
+++ b/packages/node/src/BacktraceClient.ts
@@ -8,6 +8,7 @@ import {
     BacktraceConfiguration as CoreConfiguration,
     DebugIdContainer,
     FileSystem,
+    SessionFiles,
     VariableDebugIdMapProvider,
 } from '@backtrace-labs/sdk-core';
 import path from 'path';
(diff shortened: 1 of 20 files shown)


packages/sdk-core/src/modules/breadcrumbs/BreadcrumbsManager.ts
```diff
@@ -63,11 +65,16 @@ export class BreadcrumbsManager implements BacktraceBreadcrumbs, BacktraceModule
...
+            [this.BREADCRUMB_ATTRIBUTE_NAME]: this._storage.lastBreadcrumbId,
         };
     }
 
-    public bind({ reportEvents }: BacktraceModuleBindData): void {
+    public bind({ client, reportEvents }: BacktraceModuleBindData): void {
+        for (const attachment of this._storage.getAttachments()) {
+            client.addAttachment(attachment);
+        }
+        client.addAttributeProvider(this);
```
konraddysput, line 76: qq: the AddAttribute method won't work? I think you can just use this method and do not expose additional redundant method.

packages/node/src/BacktraceClient.ts
```diff
@@ -36,19 +39,32 @@ export class BacktraceClient extends BacktraceCoreClient {
...
+            );
+        }
     }
 
     public initialize(): void {
-        super.initialize();
+        const lockId = this.sessionFiles?.lockPreviousSessions();
 
-        this.loadNodeCrashes();
+        try {
```
konraddysput, line 54: just in case - do you expect any of these method can throw an error? I think it's safe to use them without try/catch, however I also know why you added it here. Maybe it's worth to leave a note about it?

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -0,0 +1,77 @@
...
+export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
+    public get lastBreadcrumbId(): number {
+        return this._lastBreadcrumbId;
+    }
+
+    private _lastBreadcrumbId: number = TimeHelper.toTimestampInSec(TimeHelper.now());
+
+    private readonly _writer: AlternatingFileWriter;
+
+    constructor(private readonly _file1: string, private readonly _file2: string, maximumBreadcrumbs: number) {
```
konraddysput, line 27: how about renaming them to:
_file1 => _sourceFile
_file2 => _backupFile

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -0,0 +1,77 @@
...
+        this._writer = new AlternatingFileWriter(_file1, _file2, maximumBreadcrumbs);
+    }
+
+    public static createFromSession(session: SessionFiles): FileBreadcrumbsStorage | undefined {
+        const files = session
+            .getSessionFiles()
+            .filter((f) => path.basename(f).startsWith(FILE_PREFIX))
+            .slice(0, 2);
+
+        if (!files.length) {
```
konraddysput, line 37: what does it mean? There is no breadcrumb file for the current session?  If it's unexpected, shouldn't be louder about that?

packages/node/src/common/AlternatingFileWriter.ts
```diff
@@ -0,0 +1,46 @@
...
+            throw new Error('This instance has been disposed.');
+        }
+
+        if (!this._fileStream) {
+            this._fileStream = fs.createWriteStream(this._mainFile, 'utf-8');
+        } else if (this._count >= this._fileCapacity) {
+            this._fileStream.close();
+            await fs.promises.rename(this._mainFile, this._fallbackFile);
+            this._count = 0;
+            this._fileStream = fs.createWriteStream(this._mainFile);
```
konraddysput, line 29: I think it should have the same settings as a write stream from line 24

packages/node/tests/common/alternatingFile.spec.ts
```diff
@@ -0,0 +1,96 @@
+import fs from 'fs';
+import path from 'path';
+import { AlternatingFileWriter } from '../../src/common/AlternatingFileWriter';
+
+function unlinkSafe(file: string) {
```
konraddysput, line 5: One more idea for a test - when you dispose BacktraceClient, can you still add breadcrumbs? WIll it throw an error?

packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
```diff
@@ -43,6 +43,12 @@ export interface EnabledBacktraceDatabaseConfiguration {
      * A crash report is generated, stored locally, and uploaded upon next start.
      */
     captureNativeCrashes?: boolean;
+
+    /**
+     * Controls how much previous session caches to preserve before sending data from previous sessions.
```
konraddysput, line 48: Maybe it's worth to mention this option only applies for session files and is not applied to database records/generated reports in the database?

packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -0,0 +1,244 @@
...
+    private readonly _locks = new Set<string>();
+    private _previousSession?: SessionFiles;
+    private _cleared = false;
+
+    constructor(
+        private readonly _fileSystem: FileSystem,
+        private readonly _directory: string,
+        public readonly sessionId: string,
+        private readonly _maxPreviousLockedSessions = 1,
+        timestamp?: number,
```
konraddysput, line 36: how about this?

```suggestion
        private readonly _timestamp: number = Date.now()
```

packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -0,0 +1,244 @@
...
+        private readonly _lockable = true,
+    ) {
+        this._timestamp = timestamp ?? Date.now();
+        this._escapedSessionId = SessionFiles.escapeFileName(sessionId);
```
konraddysput, line 40: why you prefer to use static instead of class member function?

packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    public unlock(lockId: string) {
+        this._locks.delete(lockId);
+        if (this._locks.size === 0) {
+            this._events.emit('unlocked');
+        }
+    }
+
+    private static getFileSession(file: string): FileSession | undefined {
+        const [escapedSessionId, rawTimestamp] = this.splitByOneChar(file, '_').slice(-2);
```
konraddysput, line 190: since you know the pattern (its defined in the getFileName) can we make it simpler to parse file name? `const [_, escapedSessionId, rawTimestamp] = file.split('_');` won't work?

packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -0,0 +1,244 @@
...
+            .filter((f) => f.startsWith(SESSION_MARKER_PREFIX))
+            .map((f) => SessionFiles.getFileSession(f))
+            .filter(isDefined);
+
+        const currentSessionMarker = sessionMarkers.find((s) => s.sessionId === this.sessionId);
+
+        const lastSessionMarker = directoryFiles
+            .filter((f) => f.startsWith(SESSION_MARKER_PREFIX))
+            .map((file) => SessionFiles.getFileSession(file))
+            .filter(isDefined)
```
konraddysput, line 69: You already have session markers - you get them in line 59
```suggestion
        const lastSessionMarker = sessionMarkers
```

konraddysput, line 69: You already have session markers - you get them in line 59
```suggestion
        const lastSessionMarker = sessionMarkers
```

konraddysput, line 69: You already have session markers - you get them in line 59
```suggestion
        const lastSessionMarker = sessionMarkers
```

packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -0,0 +1,244 @@
...
+        if (!directoryFiles.length) {
+            return undefined;
+        }
+
+        const sessionMarkers = directoryFiles
+            .filter((f) => f.startsWith(SESSION_MARKER_PREFIX))
+            .map((f) => SessionFiles.getFileSession(f))
+            .filter(isDefined);
+
+        const currentSessionMarker = sessionMarkers.find((s) => s.sessionId === this.sessionId);
```
konraddysput, line 64: shoudn't you check if the currentSessionMarker exists? You're doing something similar in the function below, but we can exit this function earlier

konraddysput (approved):

packages/node/src/BacktraceClient.ts
perf2711: I don't care if the method will throw an error or not. I want to always make sure that the locks are released after.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: It's not unexpected. There may be a possibility that the previous session did not create breadcrumbs at all.

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: I'm taking the 2 last elements. If there are 4 elements, your example will not work.

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: The `filter` predicate checks if timestamp is less that currentSessionMarker's only if it exists.

packages/node/tests/common/alternatingFile.spec.ts
perf2711: This is out of scope of this test and review. I only implement the `BreadcrumbsStorage`, and not change anything about adding breadcrumbs.

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: Changed

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: Previously they were used in a static scope. Removed `static`s.

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: Good idea, changed

packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
perf2711: Good idea, changed

packages/node/src/common/AlternatingFileWriter.ts
perf2711: Yeah, good catch. Changed

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: Changed to `_mainFile` and `_fallbackFile`

adamcronin42 (approved):

packages/sdk-core/src/modules/storage/SessionFiles.ts
konraddysput: Yeah but you need to iterate over n elements when you know it won't generate a valid result if the currentSessionmarker is undefined. There is no reason to do that, right?

perf2711: I guess... changed.

packages/sdk-core/src/modules/breadcrumbs/BreadcrumbsManager.ts
konraddysput: so `client.add(() => {"last.breadcrumbid": this._lastBreadcrumbid});` won't work? it will create a dynamic attribute provider anyway. Less exposed public API - less things to maintain and less things that can break

perf2711: Hmm, I guess it will. I'll change that.

PR #151, opened 2023-09-27
add NestJS library for handling NestJS errors

perf2711:
This PR adds a NestJS library with:
* module that exposes `BacktraceClient`
* `BacktraceInterceptor` which helps catch all exceptions thrown within NestJS

Detailed usage is specific in the README. See also the e2e unit tests.

Jira task: BT-679


packages/nestjs/src/backtrace.interceptor.ts
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        return true;
+    }
+
+    private getAttributes(context: ExecutionContext) {
+        const type = context.getType();
+        switch (type) {
+            case 'http':
+                return this.getHttpAttributes(context.switchToHttp());
```
konraddysput, line 113: can we extract anything interesting from those attributes? Like for example HTTP request URL, HTTP method, controller name, etc?

packages/nestjs/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "extends": "../../tsconfig.base.json",
+    "compilerOptions": {
+        "outDir": "./lib",
+        "rootDir": "./src",
+        "composite": true
+    },
+    "exclude": ["node_modules", "lib"],
+    "references": [
+        {
+            "path": "../sdk-core/tsconfig.json"
```
konraddysput, line 11: I think you should use nodejs path instead

packages/nestjs/tsconfig.build.json
```diff
@@ -0,0 +1,14 @@
...
+    "compilerOptions": {
+        "outDir": "./lib",
+        "rootDir": "./src",
+        "composite": true
+    },
+    "exclude": ["node_modules", "tests", "lib"],
+    "references": [
+        {
+            "path": "../sdk-core/tsconfig.json"
+        }
```
konraddysput, line 12: I think you should use nodejs path instead

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -178,6 +178,13 @@ export abstract class BacktraceCoreClient {
         this._enabled = true;
     }
 
+    /**
+     * Uses client as the global client.
+     */
+    public static use(client: BacktraceCoreClient) {
```
konraddysput, line 184: We should control that, we shouldn't expose this option outside. If someone uses it and re-initialize client, we might have a huge issue with our native integrations and also with our callback functions (like breadcrumbs etc). The Backtrace client instance is protected. Each client can use it. Why do we need this?

package.json
```diff
@@ -15,8 +15,16 @@
         "test": "npm run test --ws --if-present"
     },
     "workspaces": [
-        "packages/*",
-        "tools/*"
+        "packages/sdk-core",
```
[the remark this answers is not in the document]
perf2711: I remember you doing something about this, but isn't this on the React Native branch?

https://github.com/backtrace-labs/backtrace-javascript/blob/dev/package.json#L18

packages/nestjs/src/backtrace.interceptor.ts
```diff
@@ -0,0 +1,155 @@
...
+    constructor(options: BacktraceInterceptorOptions | undefined, client: BacktraceClient);
+    constructor(
+        private readonly _options = BacktraceInterceptor.getDefaultOptions(),
+        @Optional() client?: BacktraceClient,
+    ) {
+        this._client = client;
+    }
+
+    public intercept(context: ExecutionContext, next: CallHandler<unknown>): Observable<unknown> {
+        const client = this._client ?? BacktraceClient.instance;
```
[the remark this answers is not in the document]
perf2711: Initially I did this the way you're proposing, but the client may not be initialized at moment of construction.

package.json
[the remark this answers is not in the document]
konraddysput: ugh right!

packages/nestjs/src/backtrace.interceptor.ts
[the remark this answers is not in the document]
konraddysput: Shouldn't we require to initialize the client first and later use everything else? It sounds like a natural way to use us

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: Removed the method and changed how the module is initialized.

packages/nestjs/src/backtrace.interceptor.ts
perf2711: done

konraddysput (approved):

adamcronin42 (approved):

PR #152, opened 2023-09-28
node: attributes from previous session in native crashes

perf2711:
This PR adds attributes from previous session in native crashes:
* if database and native crashes are enabled, the attributes will be saved to disk,
* every time the scoped attributes change, attributes will be resaved,
* when loading native crashes, attributes from previous session will be also loaded.

Jira task: BT-586

### DIFF
--- a/packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
+++ b/packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
@@ -1,4 +1,5 @@
 import { jsonEscaper } from '../../common/jsonEscaper';
+import { AttributeType } from '../../model/data';
 import { BacktraceModule, BacktraceModuleBindData } from '../BacktraceModule';
 import { FileSystem, SessionFiles } from '../storage';
 import { AttributeManager } from './AttributeManager';
@@ -23,7 +24,7 @@ export class FileAttributeManager implements BacktraceModule {
         this.saveAttributes();
     }
 
-    public bind?({ attributeManager, sessionFiles }: BacktraceModuleBindData): void {
+    public bind({ attributeManager, sessionFiles }: BacktraceModuleBindData): void {
         if (this._fileName) {
             throw new Error('This instance is already bound.');
         }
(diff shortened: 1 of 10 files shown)


packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
```diff
@@ -0,0 +1,65 @@
...
+    public static createFromSession(sessionFiles: SessionFiles, fileSystem: FileSystem) {
+        const fileName = sessionFiles.getFileName(ATTRIBUTE_FILE_NAME);
+        return new FileAttributeManager(fileSystem, fileName);
+    }
+
+    public initialize(): void {
+        this.saveAttributes();
+    }
+
+    public bind?({ attributeManager, sessionFiles }: BacktraceModuleBindData): void {
```
konraddysput, line 26:
```suggestion
    public bind({ attributeManager, sessionFiles }: BacktraceModuleBindData): void {
```

packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
```diff
@@ -0,0 +1,65 @@
...
+        this._fileName = sessionFiles.getFileName(ATTRIBUTE_FILE_NAME);
+        this._attributeManager = attributeManager;
+        attributeManager.attributeEvents.on('scoped-attributes-updated', () => this.saveAttributes());
+    }
+
+    public dispose?(): void {
+        this._fileName = undefined;
+    }
+
+    public async get(): Promise<Record<string, unknown>> {
```
konraddysput, line 44: I think you can use AttributeType instead of unknown. 
```suggestion
    public async get(): Promise<Record<string, AttributeType>> {
```

konraddysput (approved):

packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
perf2711: Done

packages/sdk-core/src/modules/attribute/FileAttributeManager.ts
perf2711: Done

adamcronin42 (approved):

PR #153, opened 2023-09-29
Do not use unsupported URL API in the react-native API

konraddysput:
# Why

React-Native cannot use the URL API. Unfortunately, most of the method available in the URL interface is unavailable (except some of them). Since the SubmissionUrl API is easy to implement manually, we can avoid using URL and use manual string manipulation to construct the URL.

This diff adjust the SubmissionUrl class to be re-usable in all our libraries

### DIFF
--- a/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
+++ b/packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
@@ -48,7 +48,7 @@ export class SubmissionUrlInformation {
 
     /**
      * Find the universe based on the submission URL
-     * @param submissionUrl submission URL
+     * @param submissionUrl submission URL - full submission URL to Backtrace.
      * @returns universe name
      */
     public static findUniverse(submissionUrl: string): string | undefined {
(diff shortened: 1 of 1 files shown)


packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -80,7 +89,7 @@ export class SubmissionUrlInformation {
             // submit.backtrace.io/universe/token/format
             // by spliting the submission URL by `/` and dropping the last
             // part of the URL, the last element on the list is the token.
-            return submissionUrlParts[submissionUrlParts.length - 2];
+            return submissionUrlParts[submissionUrlParts.length - 2] ?? null;
         }
 
         const url = new URL(submissionUrl);
```
perf2711, line 95: You still use `URL` here.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
```diff
@@ -90,22 +99,33 @@ export class SubmissionUrlInformation {
...
-            pathParts[3] = desiredFormat;
-            url.pathname = pathParts.join('/');
+            pathParts[2] = desiredFormat;
+            return (
+                submissionUrl.substring(0, submitIndex + this.SUBMIT_PREFIX.length) +
+                pathParts.join('/') +
+                queryParameters
+            );
         } else {
+            const url = new URL(submissionUrl);
```
perf2711, line 126: You still use `URL` here.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
konraddysput: As I said - we're using only supported URL API. If something is unsupported I'm creating the URL manually.

packages/sdk-core/src/model/http/SubmissionUrlInformation.ts
konraddysput: As I said - we're using only supported URL API. If something is unsupported I'm creating the URL manually.

perf2711 (approved):

PR #162, opened 2023-10-06
add methods to BacktraceBreadcrumbs interface and limit BacktraceManager visibility

perf2711:
This PR changes the following things:
* changes `BacktraceManager` to `BacktraceBreadcrumbs` in `start` in `BacktraceEventSubscriber`,
* add `addBreadcrumb` method to `BacktraceBreadcrumbs`,
* add `addEventSubscriber` to `BacktraceBreadcrumbs` and `BacktraceManager`,
* set `_enabled = true` in `BreadcrumbsManager` after `initialize`


packages/browser/src/breadcrumbs/DocumentEventSubscriber.ts
```diff
@@ -1,14 +1,14 @@
...
     BreadcrumbsEventSubscriber,
-    BreadcrumbsManager,
     BreadcrumbType,
 } from '@backtrace-labs/sdk-core';
 
 export class DocumentEventSubscriber implements BreadcrumbsEventSubscriber {
     private readonly _controller: AbortController = new AbortController();
 
-    public start(breadcrumbsManager: BreadcrumbsManager): void {
+    public start(breadcrumbsManager: BacktraceBreadcrumbs): void {
```
konraddysput, line 11: I think you should also rename the variable

konraddysput (changes requested):
Can we rename `breadcrumbsManager` to `backtraceBreadcrumbs` or something more similar for the new interface?

packages/browser/src/breadcrumbs/DocumentEventSubscriber.ts
perf2711: Done

konraddysput (approved):

PR #163, opened 2023-10-06
sync versions script

perf2711:
This script helps to synchronize all versions of workspace packages in all workspaces.

For example, if `sdk-core` version changes, instead of going to every package and updating the version manually, one has to only run `npm run syncVersions` and it will take care of everything.

Example:
1. update sdk-core from 0.0.5 to 0.0.6
2. run `npm run syncVersions`
Output:
```
> @backtrace-labs/javascript@0.0.1 syncVersions
> ts-node ./scripts/syncVersions.ts

[@backtrace-labs/sdk-core] - no changes
[@backtrace-labs/browser] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
[@backtrace-labs/react] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
[@backtrace-labs/node] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
[@backtrace-labs/nestjs] - no changes
[@backtrace-labs/react-native] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
[@backtrace-labs/sourcemap-tools] - no changes
[@backtrace-labs/javascript-cli] - no changes
[@backtrace-labs/rollup-plugin] - no changes
[@backtrace-labs/webpack-plugin] - no changes
[@backtrace-labs/vite-plugin] - no changes
```

Also, if you want to sync versions of specific packages, you can pass paths to package.jsons:
```
> npm run syncVersions -- packages/node/package.json packages/browser/package.json 

> @backtrace-labs/javascript@0.0.1 syncVersions
> ts-node ./scripts/syncVersions.ts packages/node/package.json packages/browser/package.json

[@backtrace-labs/node] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
[@backtrace-labs/browser] - updated @backtrace-labs/sdk-core from ^0.0.5 to ^0.0.6 in dependencies
```

### DIFF
--- a/scripts/syncVersions.ts
+++ b/scripts/syncVersions.ts
@@ -47,36 +47,36 @@ function getWorkspacePackageJsonPaths(packageJson: PackageJson) {
     return packageJson.workspaces?.map((workspaceDir) => path.join(rootDir, workspaceDir, 'package.json')) ?? [];
 }
 
+function updateDependency(packageJson: PackageJson, type: Dependencies, name: string, newVersion: string) {
+    const deps = packageJson[type];
+    if (!deps) {
+        return false;
+    }
+
+    const currentVersion = deps[name];
+    if (!currentVersion) {
+        return false;
+    }
+
+    if (currentVersion !== newVersion) {
+        deps[name] = newVersion;
+        console.log(`[${packageJson.name}] - updated ${name} from ${currentVersion} to ${newVersion} in ${type}`);
+
+        return true;
+    }
+
+    return false;
+}
+
 function updateVersions(packageJson: PackageJson, currentVersions: Record<string, string>) {
     let updated = false;
 
     for (const [name, version] of Object.entries(currentVersions)) {
         const newVersion = `^${version}`;
 
-        function updateDependency(type: Dependencies) {
-            const deps = packageJson[type];
-            if (!deps) {
-                return;
-            }
-
-            const currentVersion = deps[name];
-            if (!currentVersion) {
-                return;
-            }
-
-            if (currentVersion !== newVersion) {
-                deps[name] = newVersion;
-                console.log(
-                    `[${packageJson.name}] - updated ${name} from ${currentVersion} to ${newVersion} in ${type}`,
-                );
-
-                return true;
-            }
-        }
-
-        updated = updateDependency('dependencies') || updated;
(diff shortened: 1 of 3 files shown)


scripts/syncVersions.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const args = process.argv.slice(2);
+    const packageJsonPathsToSync = args.length ? args : workspacePackageJsonPaths;
+
+    const currentVersions = workspacePackageJsons.reduce((obj, pj) => {
+        obj[pj.name] = pj.version;
+        return obj;
+    }, {} as Record<string, string>);
+
+    for (const packageJsonPath of packageJsonPathsToSync) {
```
konraddysput, line 102: can we run `npm install` in the end if there is any change? 

we're using `npm ci` which requires from us to keep our package-lock.json up to date. We can also automate this by using this tool. 

WDYT?

scripts/syncVersions.ts
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+    }
+
+    const currentVersion = deps[name];
+    if (!currentVersion) {
+        return false;
+    }
+
+    if (currentVersion !== newVersion) {
+        deps[name] = newVersion;
```
konraddysput, line 62: shouldn't this be `^${newVersion}` instead? Otherwise our concept with sdk-core won't work

konraddysput (commented):
Left some comments before we land the tool. WDYT?

scripts/syncVersions.ts
perf2711: it is, see line 75

scripts/syncVersions.ts
perf2711: I wouldn't do that. Even if we decide that, it should be a part of scripts in package.json: 
```
ts-node ./scripts/syncVersions.ts && npm install
```

scripts/syncVersions.ts
konraddysput: ah ok. thx

konraddysput (approved):

adamcronin42 (approved):

PR #167, opened 2023-10-09
node: Remove overflowed records on the database startup and replace foreach with for to iterate over all records rather than jump between them

konraddysput:
# Why

Database send (interval method) was using the foreach loop instead of for. By doing that, removing a record and iterating into the next one, skipped some records

### DIFF
--- a/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
+++ b/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
@@ -198,7 +198,11 @@ export class BacktraceDatabase implements BacktraceModule {
      */
     public async send() {
         for (let bucketIndex = 0; bucketIndex < this._databaseRecordContext.bucketCount; bucketIndex++) {
-            for (const record of this._databaseRecordContext.getBucket(bucketIndex)) {
+            // make a copy of records to not update the array after each remove
+            const records = [...this._databaseRecordContext.getBucket(bucketIndex)];
+
+            for (let recordIndex = 0; recordIndex < records.length; recordIndex++) {
+                const record = records[recordIndex];
                 if (record.locked) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -198,7 +198,11 @@ export class BacktraceDatabase implements BacktraceModule {
      */
     public async send() {
         for (let bucketIndex = 0; bucketIndex < this._databaseRecordContext.bucketCount; bucketIndex++) {
-            for (const record of this._databaseRecordContext.getBucket(bucketIndex)) {
+            // make a copy of records to not update the array after each remove
+            const records = [...this._databaseRecordContext.getBucket(bucketIndex)];
+
+            for (let recordIndex = 0; recordIndex < records.length; recordIndex++) {
+                const record = records[recordIndex];
```
perf2711, line 205: We can surely keep `foreach` if we are copying the array?

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -246,6 +249,12 @@ export class BacktraceDatabase implements BacktraceModule {
         }
     }
 
+    private deleteRecords(recordsToDelete: BacktraceDatabaseRecord[]) {
+        for (const recordToDelete of recordsToDelete) {
+            this._storageProvider.delete(recordToDelete);
```
perf2711, line 254: Why not call `remove` here?

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
konraddysput: Fair - My goal was to avoid foreach overall in the database operations, but we can keep the code simple. Changing to foreach

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
konraddysput: I forgot about remove - changing to remove

perf2711 (approved):

PR #170, opened 2023-10-09
nestjs: version 0.0.2

perf2711:


packages/nestjs/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@backtrace-labs/nestjs",
-    "version": "0.0.1",
+    "version": "0.0.2",
```
konraddysput, line 3: we havent released the library yet. I do not see the lib on npm

konraddysput (changes requested):

PR #173, opened 2023-10-10
sdk-core, browser, node, react, react-native: builder changes

perf2711:
This PR changes the builder, clients and its inheritors to:
* accept only `CoreClientSetup` or extending interfaces,
* set defaults in the client constructors as much as possible,
* add generic arguments for setup/configurations.

This change makes it easier to determine where the default for some implementation is set. Also, it enables to add stuff to core builder without having to make changes to inheriting builders.


packages/sdk-core/src/builder/BacktraceCoreClientBuilder.ts
```diff
@@ -1,51 +1,56 @@
...
         return this;
     }
 
     public useFileSystem(fileSystem: FileSystem): this {
-        this.fileSystem = fileSystem;
+        this.clientSetup.fileSystem = fileSystem;
         return this;
     }
-
-    public abstract build(): T;
```
perf2711, line 50: Removal of this method helps us to get rid of `T` argument. This method is not used anywhere in an abstract context.

packages/browser/src/builder/BacktraceClientSetup.ts
```diff
@@ -0,0 +1,5 @@
+import { PartialCoreClientSetup } from '@backtrace-labs/sdk-core';
+import { BacktraceConfiguration } from '../BacktraceConfiguration';
+
+export interface BacktraceClientSetup<O extends BacktraceConfiguration = BacktraceConfiguration>
+    extends PartialCoreClientSetup<'sdkOptions' | 'requestHandler', O> {}
```
konraddysput, line 5: we need to provide an option to override requestHandler. I think the change removes requestHandler from the browser package options. Can we bring it?

packages/browser/src/index.ts
```diff
@@ -10,17 +10,19 @@ export {
     BacktraceStringAttachment,
     BacktraceUint8ArrayAttachment,
     BreadcrumbLogLevel,
+    BreadcrumbType,
```
konraddysput, line 13: I use alphabetical sorting for usings/imports in VSC. Is there any reason to change that? What is your config, it looks odd vsc ordered it this way

packages/react-native/src/builder/BacktraceClientBuilder.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Platform } from 'react-native';
+import { BacktraceClient } from '../BacktraceClient';
+import { NativeAttributeProvider } from '../attributes/NativeAttributeProvider';
+import type { BacktraceClientSetup } from './BacktraceClientSetup';
+
+export class BacktraceClientBuilder extends BacktraceCoreClientBuilder<BacktraceClientSetup> {
+    constructor(clientSetup: BacktraceClientSetup) {
+        super(clientSetup);
+
+        const attributeProviders = Platform.select({
```
konraddysput, line 11: 1. We need to verify if the debugger is connected. Otherwise, attributes might break react-native debugging experience in Chrome. 
2. You're also missing ReactNativeAttributeProvider

konraddysput (changes requested):
Our clients have a case where they want to override our HTTP client. We need to expose request handelr interface and let them override it.

In addition to that the react-native builder is broken - can you add a few more checks (you can use my code from the builder in your new class)

packages/browser/src/index.ts
perf2711: well, `BreadcrumbsManager` is after `BreadcrumbType` alphabetically

packages/react-native/src/builder/BacktraceClientBuilder.ts
perf2711: good catch, I've missed those. I added them back.

packages/browser/src/index.ts
perf2711: wait, it's not 😄 

My VSC also sorts this automatically on save.

image

perf2711: it seems that for me it's case-sensitive, and 'T' is before 's' as a char code.
image

adamcronin42 (approved):

PR #176, opened 2023-10-16
electron: SDK for Electron

perf2711:
The Electron SDK is finally here!

This PR adds a working Backtrace Electron integration.
* Main process uses Electron `BacktraceClient` which is based on node.
* Renderer process can use any `BacktraceClient` derived from browser.
* Submission process is done by main process entirely. Renderer client has overrided behavior to pass all data to main.
* Attributes are independent in main and renderer, with the exception of sessionId, which is taken from main.
* Breadcrumbs are joined together, and main is storing them all.
* Attachments are joined and sent together from main, and transferred via streams from renderer to main.
* Renderer attachments are currently NOT included when an error occurs in main, but they are included when an error occurs in renderer.
* Crashpad is used for native crashes. The data is stored in the database folder.

This release is NOT production ready - docs need to be written, READMEs, better import behavior, etc.

Jira task: BT-709


packages/electron/src/common/ipc/IpcRpc.ts
```diff
@@ -0,0 +1,25 @@
+import { Event } from 'electron';
+
+export interface IpcRpcCaller {
+    invoke<T>(event: string, ...args: unknown[]): Promise<T>;
+}
+
+export interface IpcRpcHandler {
+    on(event: string, callback: (event: Event, ...args: any[]) => Promise<any>): this;
```
konraddysput, line 8: can we please ignore any if we want to keep it? Please take a look into Github warnings

packages/electron/src/renderer/addBacktraceElectron.ts
```diff
@@ -0,0 +1,53 @@
...
+    builder
+        .useRequestHandler(new IpcRequestHandler(ipcRpc))
+        .useReportSubmission(new IpcReportSubmission(ipcRpc, ipcTransport))
+        .useBreadcrumbsStorage(new IpcBreadcrumbsStorage(ipcTransport))
+        .useSummedMetricsQueue(new IpcSummedMetricsQueue(ipcTransport, ipcRpc))
+        .useUniqueMetricsQueue(new StubMetricsQueue());
+
+    if (options?.synchronous) {
+        const syncData = ipcRpc.invokeSync<SyncData>(IpcEvents.sync);
+        console.log(syncData);
```
konraddysput, line 45: console.log

packages/electron/src/renderer/addBacktraceElectron.ts
```diff
@@ -0,0 +1,53 @@
...
+        builder.useSessionProvider(new ConstSessionProvider(syncData.sessionId));
+    } else {
+        builder.useSessionProvider(new IpcAsyncSessionProvider(ipcTransport));
```
konraddysput, line 48: is there any reason anyone should prefer to have it instead of sync? 

The nature of programming is async all the way, however, it doesn't work well for error reporting solutions, while we're waiting on something, an error might happen. Can we stick to the sync session provider and introduce the async interface once we recognize we have to do something asynchronously? There is no real benefit other than "async all the way all the time without any good reason".

packages/electron/src/renderer/addBacktraceElectron.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IpcReportSubmission } from './modules/IpcReportSubmission';
+import { IpcRequestHandler } from './modules/IpcRequestHandler';
+import { IpcSummedMetricsQueue } from './modules/IpcSummedMetricsQueue';
+import { StubMetricsQueue } from './modules/StubMetricsQueue';
+
+export interface AddBacktraceElectronOptions {
+    readonly synchronous?: boolean;
+}
+
+export function addBacktraceElectron<T extends BacktraceCoreClientBuilder>(
```
konraddysput, line 18: Can we add documentation what it does any why someone should use it?

packages/electron/src/renderer/addBacktraceElectron.ts
```diff
@@ -0,0 +1,53 @@
...
+    options?: AddBacktraceElectronOptions,
+): T {
+    const ipcTransport = getIpcTransport();
+    const ipcRpc = getIpcRpc();
+
+    // Sanity check - this will throw if ping isn't listened on.
+    ipcRpc
+        .invoke(IpcEvents.ping)
+        .catch((err) =>
+            console.error(
```
konraddysput, line 29: should the integration be disabled? Right now the builder will build the client in a way to forward everything to the main process, so I'm assuming it won't work and it will probably generate an infinity loop of errors?

packages/electron/src/main/BacktraceClient.ts
```diff
@@ -0,0 +1,40 @@
...
+    BacktraceSetupConfiguration,
+    BacktraceClient as NodeBacktraceClient,
+} from '@backtrace-labs/node';
+import { BacktraceClientBuilder } from './builder/BacktraceClientBuilder';
+import { BacktraceMainElectronModule } from './modules/BacktraceMainElectronModule';
+
+export class BacktraceClient extends NodeBacktraceClient {
+    constructor(clientSetup: BacktraceNodeClientSetup) {
+        super({
+            ...clientSetup,
```
konraddysput, line 12: I think it's important to also use electron API to get application name and application version. In node we will try to understand the app structure and verify env, but it most likely won't work. 

https://www.electronjs.org/docs/latest/api/app#appgetversion

packages/electron/src/main/builder/BacktraceClientBuilder.ts
```diff
@@ -0,0 +1,10 @@
+import { BacktraceNodeClientSetup, BacktraceClientBuilder as NodeBacktraceClientBuilder } from '@backtrace-labs/node';
+import { BacktraceClient } from '../BacktraceClient';
+
+export class BacktraceClientBuilder extends NodeBacktraceClientBuilder {
```
konraddysput, line 4: can we provide more attribute providers? 

We can get at least:
- application name
- application version
- OS culture
- app metrics 
- GPU information
- check if electron is running in the unity env
- is application packaged?
- window information like fullscreen, window size, kiosk mode etc
- information if the app is online


Definitely the app might not behave correctly if we package it because of lack of application name/version that we should expose via electron - node can fail here

konraddysput (changes requested):

packages/electron/src/renderer/addBacktraceElectron.ts
perf2711: The reason is that if you misconfigure something, there's no real way to tell the user "you screwed up" with synchronous code. It will just hang with a white screen.

I believe this may be counterintuitive, and may lead to the user removing Backtrace from his app and then seeing that removing fixed the white screen issue.

packages/electron/src/renderer/addBacktraceElectron.ts
perf2711: I'll make sure to add that in the future.

As stated in the PR description:
>This release is NOT production ready - docs need to be written, READMEs, better import behavior, etc.

packages/electron/src/renderer/addBacktraceElectron.ts
perf2711: There's no way for me to disable the integration, and to throw this error synchronously. Least I could do is to display a big error message.

konraddysput: client.disable won't work?

perf2711: This is at the building stage, the client is not built yet.

packages/electron/src/renderer/addBacktraceElectron.ts
perf2711: removed

packages/electron/src/common/ipc/IpcRpc.ts
perf2711: changed possible `any` to `unknown` and added skips for the left

packages/electron/src/main/BacktraceClient.ts
perf2711: Done!

packages/electron/src/main/builder/BacktraceClientBuilder.ts
perf2711: Added additional providers.

packages/electron/src/main/attributes/ReadyAppAttributeProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    public readonly type: 'scoped' | 'dynamic';
+
+    constructor() {
+        if (app.isReady()) {
+            this.type = 'scoped';
+            this._attributes = this.buildAttributes();
+        } else {
+            this.type = 'dynamic';
+            app.on('ready', () => (this._attributes = this.buildAttributes()));
```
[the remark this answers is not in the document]
perf2711: yes

packages/electron/src/renderer/addBacktraceElectron.ts
konraddysput: so maybe we should always stick to the async one and update it in the future? Tbh if we update it as soon as possible, I believe it won't be a problem

PR #177, opened 2023-10-16
electron example: add an example for using Electron SDK

perf2711:
This PR adds an example showcasing features of Backtrace in Electron, and how to initialize it in an Electron app.

Jira task: BT-709

### DIFF
--- a/examples/sdk/electron/assets/index.html
+++ b/examples/sdk/electron/assets/index.html
@@ -1,0 +1,115 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html lang="en">
+    <head>
+        <link rel="preconnect" href="https://fonts.googleapis.com" />
+        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="" />
+        <link href="https://fonts.googleapis.com/css2?family=DM+Sans:wght@400;700&display=swap" rel="stylesheet" />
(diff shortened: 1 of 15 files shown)


examples/sdk/electron/assets/index.html
```diff
@@ -0,0 +1,140 @@
...
+                            <span class="text">Generate metric</span>
+                        </div>
+                        <div class="action-button generate-metric" data-target="main">
+                            <a class="text" target="_blank">Main</a>
+                        </div>
+                        <div class="action-button generate-metric" data-target="renderer">
+                            <a class="text" target="_blank">Renderer</a>
+                        </div>
+                    </div>
+                    <div class="button-container center">
```
konraddysput, line 83: Can we add crash?

examples/sdk/electron/src/main/consts.ts
```diff
@@ -0,0 +1,2 @@
+export const SUBMISSION_URL =
+    'https://salex.in.backtrace.io:6098/post?format=json&token=81ebe51f9caa641c2013b15b59f704a853a70e542ce279a90bf1526853d8ff0a';
```
konraddysput, line 2: salex submission url

konraddysput (changes requested):

examples/sdk/electron/assets/index.html
perf2711: 94-96:
```
<div class="action-button crash-app" data-target="main">
                        <a class="text" target="_blank">Crash application</a>
                    </div>
```

konraddysput (approved):

PR #178, opened 2023-10-16
react-native: Native database ios

konraddysput:
# Why

This diff adds support for file operations in the database in iOS. All database operations are supported.


packages/react-native/android/src/main/java/backtrace/library/BacktraceFileSystemProvider.java
```diff
@@ -137,16 +134,26 @@ public void exists(String path, Promise promise) {
...
+    }
+
+    private boolean copy(String sourcePath, String destinationPath) {
+        try (FileInputStream inputStream = new FileInputStream(sourcePath);
+             FileOutputStream outputStream = new FileOutputStream(destinationPath, false)) {
+            byte[] buf = new byte[1024];
+            int len;
+            while ((len = inputStream.read(buf)) > 0) {
+                outputStream.write(buf, 0, len);
+            }
```
perf2711, line 153: What Java version do we use here?
You could use `transferTo`: https://docs.oracle.com/javase/9/docs/api/java/io/InputStream.html#transferTo-java.io.OutputStream-

perf2711 (approved):

packages/react-native/android/src/main/java/backtrace/library/BacktraceFileSystemProvider.java
konraddysput: image

PR #185, opened 2023-10-20
electron: add docs and README

perf2711:


packages/electron/src/renderer/addBacktraceElectron.ts
```diff
@@ -12,9 +12,27 @@ import { IpcSummedMetricsQueue } from './modules/IpcSummedMetricsQueue';
 import { StubMetricsQueue } from './modules/StubMetricsQueue';
 
 export interface AddBacktraceElectronOptions {
+    /**
+     * Data needs to be synchronized with the main process.
+     * If this is set to `true`, this synchronization will be synchronous and blocking.
+     * This can help with startup crashes not having the whole information, like session ID.
+     *
+     * **If Backtrace is not fully initialized on the main process, your application may hang with a white screen**.
```
konraddysput, line 20: What is the default value?

konraddysput (changes requested):
Can we add a default value?

packages/electron/src/renderer/addBacktraceElectron.ts
perf2711: `false`. Added.

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+addBacktraceElectron(builder);
+
+const client = builder.build();
+```
+
+#### Synchronized initialization
+
+When adding Electron to the renderer SDK, data is exchanged with the main process to ensure same session IDs are used.
+By default, this is done asynchronously as soon as possible. This may however be not enough in case of startup crashes
```
konraddysput, line 121: every error will go thru the main process anyway. So the application session shouldn't be different, am I correct? Also crashes always go thru the main process so the application.session is always there.

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+just after the SDK initialization, because the data may not be exchanged yet.
+
+In those cases, you can use `synchronize` option in `addBacktraceBuilder`. Note, that if you do not ensure that the SDK
```
konraddysput, line 124: actually, the only reason why synchronize could be cool is to make sure the application.session attribute is always the same as it is in the main process. Do we do anything else? I think we don't mention that here.

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+    {
+        url: '',
+        name: 'My Renderer App',
+        version: '0.0.1',
+    },
+    (builder) => addBacktraceElectron(builder, { synchronous: true }),
+);
+```
+
+### Upload source maps
```
konraddysput, line 142: I think we also need to add information about crash symbolication. I think we can do it outside this pull request but I'm pointing this out to not forget about this

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+## Error Reporting Features
+
+### Attributes
+
+Custom attributes are key-value pairs that can be added to your error reports. They are used in report aggregation,
+sorting and filtering, can provide better contextual data for an error, and much more. They are foundational to many of
+the advanced Backtrace features detailed in
+[Error Reporting documentation](https://docs.saucelabs.com/error-reporting/getting-started/). By default attributes such
+as application name and version are populated automatically based on your package.json information. If Backtrace cannot
```
konraddysput, line 159: we populate them based on the electron application name/version. i think we can remove this sentence from here. It's important to note it in node, however in electron, we have an obvious way to read those info

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+find them, you need to provide them manually via userAttributes attributes.
+
+There are several places where attributes can be added, modified or deleted.
+
+#### Attach attributes object to BacktraceClient
+
+It is possible to include an attributes object during [BacktraceClient](#backtraceclient) initialization. This list of
+attributes will be included with every error report, referred to as global attributes.
+
+```ts
```
konraddysput, line 169: should we point this is the main process?

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+---
+
+### File Attachments
+
+Files can be attached to error reports. This can be done when initalizing the BacktraceClient, updating the
+BacktraceClient, or dynamically for specific reports. When including attachments in BacktraceClient, all files will be
+uploaded with each report.
+
+```ts
+// Import attachment types from @backtrace/electron
```
konraddysput, line 246: should it be commented?

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+report.attributes['myReportKey'] = 'New value';
+```
+
+---
+
+### File Attachments
+
+Files can be attached to error reports. This can be done when initalizing the BacktraceClient, updating the
+BacktraceClient, or dynamically for specific reports. When including attachments in BacktraceClient, all files will be
+uploaded with each report.
```
konraddysput, line 243: I think its worth to mention we support file attachments in the renderer process - or actually do we? How we can pass file attachments from renderer?

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+```ts
+// Import attachment types from @backtrace/electron
+import { BacktraceStringAttachment, BacktraceUint8ArrayAttachment, BacktraceFileAttachment } from "@backtrace/electron";
+
+// BacktraceStringAttachment should be used for text object like a log file, for example
+const stringAttachment = new BacktraceStringAttachment("logfile.txt", "This is the start of my log")
+
+// Buffer attachment is an attachment type dedicated to store buffer data
+const bufferAttachment = new BacktraceBufferAttachment('buffer-attachment.txt', Buffer.from('sample'));
```
konraddysput, line 253: Buffer attachment is not present in the import statement

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+```
+
+---
+
+### Offline database support
+
+The Backtrace Electron SDK can cache generated reports and crashes to local disk before sending them to Backtrace. This
+is recommended; in certain configurations Electron applications can crash before the SDK finishes submitting data, and
+under slow internet conditions your application might wait in a closing window until the HTTP submission finishes. In
+such an event cached reports will be sent on next application launch.
```
konraddysput, line 368: It's not true, Crashpad is out of process crash reporting solution so we report reports immediately on crash or when we fail we report it on app restart

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+```ts
+// metrics will be undefined if not enabled
+client.metrics?.send();
+```
+
+---
+
+### Offline database support
+
+The Backtrace Electron SDK can cache generated reports and crashes to local disk before sending them to Backtrace. This
```
konraddysput, line 365: I think it's worth to mention we're using crash reporting solution available in electron. If the user overrides it, crashes might not be available in Backtrace

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+| `maximumNumberOfRecords`  | Number  | The maximum number of reports stored in the offline database. When the limit is reached, the oldest reports are removed. If the value is equal to '0', then no limit is set. | 8       |
+| `retryInterval`           | Number  | The amount of time (in ms) to wait between retries if the database is unable to send a report.                                                                               | 60 000  |
+| `maximumRetries`          | Number  | The maximum number of retries to attempt if the database is unable to send a report.                                                                                         | 3       |
+| `captureNativeCrashes`    | Boolean | Capture and symbolicate stack traces for native crashes if the runtime supports this. A crash report is generated, stored locally, and uploaded upon next start.             | false   |
+
+#### Native crash support
+
+The Backtrace Electron SDK can capture native crashes generated by a Electron application such as Assert/OOM crashes. In
```
konraddysput, line 410: We do collect all native crashes. Not only OOMs/Asserts. TO be honest, if we don't have enough luck, we won't collect ooms, or they will be visible as a regular crash (failed to allocate memory or something similar)

konraddysput (changes requested):
Can we check readme once again?

packages/electron/README.md
perf2711: There are other things potentially using the session ID.

packages/electron/README.md
perf2711: There are other things potentially using the session ID.

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+1. URL is not used - you can leave it as an empty string,
+1. You need to use `addBacktraceElectron` with the builder.
+
+For example, if you are initializing the SDK using the `initialize` function:
+
+```ts
+const client = BacktraceClient.initialize(
+    {
+        url: '',
+        name: 'My Renderer App',
```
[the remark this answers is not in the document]
konraddysput: No action is needed here just in case, THis is just a wish. i do not have an idea what can we do here

packages/electron/README.md
perf2711: The file attachments are passed, but only when the error is thrown in the renderer. Attaching renderer attachments when an error is thrown in main is a bit more complicated, and it is not implemented here.

packages/electron/README.md
perf2711: >such as

This does not limit the crashes to these options only.

packages/electron/README.md
perf2711: `BacktraceMetrics`, for instance. `sessionId` is public there.

packages/electron/README.md
```diff
@@ -0,0 +1,504 @@
...
+client.send(new Error('Something broke!'));
+```
+
+### Integrate the SDK in the renderer process
+
+You can use any browser based Backtrace SDK in your renderer process, such as `@backtrace/browser` or
+`@backtrace/react`. Make sure to install it aside `@backtrace/electron`:
+
+```
+$ npm install @backtrace/react
```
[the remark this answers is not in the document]
perf2711: Done.

packages/electron/README.md
perf2711: Done.

packages/electron/README.md
perf2711: Done.

packages/electron/README.md
perf2711: Uh, nope?

packages/electron/README.md
perf2711: Changed.

packages/electron/README.md
perf2711: Changed.

packages/electron/README.md
perf2711: This is specified in `Native crash support`.

packages/electron/README.md
konraddysput: yeah this is what I said in my comment. Do you see any other module impacted because of lack of synchronization between main and renderer?

packages/electron/README.md
konraddysput: Can we note this here? Also if someone passes in memory attachment to the send method, does it mean we won't pass it to main?

packages/electron/README.md
konraddysput: can we remove such as? and just focus on we're capturing crashes?

packages/electron/README.md
perf2711: Metrics use `sessionId`, and have a public property exposing the session ID for instance.

I feel like I answered this before... github is losing my comments!

packages/electron/README.md
perf2711: No, I don't believe so. That may change however in the future, as multiple things may be using the session ID provider.

packages/electron/README.md
perf2711: Done

packages/electron/README.md
perf2711: Memory attachment? Generally, I try to send them anyway by serializing them to JSON.

image

Added a section in "Integrate the SDK in the renderer process" about the attachments.

konraddysput (approved):

PR #186, opened 2023-10-23
Library rename backtrace-labs -> backtrace

konraddysput:
# Why

This diff renames all library names to @backtrace from @backtrace-labs

refs BT-733


rick-bt (approved):
Yay!

perf2711 (approved):
As long as you did replace `@backtrace-labs` to `@backtrace`, it's all right.

PR #187, opened 2023-10-25
sdk-core, browser, node: add abort support to metrics and request handlers

perf2711:
This PR adds support for abort signal for metrics and request handlers.
This fixes the issue of dangling timeout handles when the client is disposed.

Also, a polyfill for `AbortController` for Node 14 is added.

### DIFF
--- a/packages/sdk-core/tests/metrics/summedEventTests.spec.ts
+++ b/packages/sdk-core/tests/metrics/summedEventTests.spec.ts
@@ -63,7 +63,7 @@ describe('Summed events tests', () => {
         expect(testHttpClient.post).toBeCalledWith(
             summedEventsSubmissionUrl,
             JSON.stringify(expectedJson),
-            expect.any(AbortSignal),
+            expect.anything(),
         );
     });
 
@@ -105,7 +105,7 @@ describe('Summed events tests', () => {
         expect(testHttpClient.post).toBeCalledWith(
             summedEventsSubmissionUrl,
             JSON.stringify(expectedJson),
-            expect.any(AbortSignal),
+            expect.anything(),
         );
     });
 
@@ -143,11 +143,7 @@ describe('Summed events tests', () => {
         metrics.initialize();
 
         expect(attributeManager.get().attributes).toMatchObject(customAttributes);
-        expect(testHttpClient.post).toBeCalledWith(
-            expect.anything(),
-            JSON.stringify(expectedJson),
-            expect.any(AbortSignal),
-        );
+        expect(testHttpClient.post).toBeCalledWith(expect.anything(), JSON.stringify(expectedJson), expect.anything());
     });
 
     it('Should add summed event to the submission queue', () => {
--- a/packages/sdk-core/tests/metrics/uniqueEventTests.spec.ts
+++ b/packages/sdk-core/tests/metrics/uniqueEventTests.spec.ts
@@ -64,7 +64,7 @@ describe('Unique events tests', () => {
         expect(testHttpClient.post).toBeCalledWith(
             uniqueEventsSubmissionUrl,
             JSON.stringify(expectedJson),
-            expect.any(AbortSignal),
+            expect.anything(),
         );
     });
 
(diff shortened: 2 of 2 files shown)


packages/sdk-core/src/common/DelayHelper.ts
```diff
@@ -6,12 +8,21 @@ export class Delay {
      */
     public static wait(timeout: number, signal?: AbortSignal) {
         return new Promise<void>((resolve, reject) => {
+            // intervalId has to be defined here,
+            // as abortCallback can execute before setTimeout is executed
+            // eslint-disable-next-line prefer-const
+            let intervalId: NodeJS.Timeout | undefined;
```
konraddysput, line 14: does it work correctly with types in the browser?  I think you can use `ReturnType<typeof setTimeout>` instead

packages/sdk-core/src/common/DelayHelper.ts
```diff
@@ -6,12 +8,21 @@ export class Delay {
...
+
             function abortCallback() {
                 clearTimeout(intervalId);
-                reject(new Error('Operation cancelled.'));
+                reject(new AbortError());
```
konraddysput, line 18: I think this can cause an unhandled promise rejection that we can throw on the app exit. Because of that the app can exit with a different exit code. Am I Correct?

packages/sdk-core/src/model/http/BacktraceRequestHandler.ts
```diff
@@ -14,12 +14,13 @@ export interface BacktraceRequestHandler {
         submissionUrl: string,
         dataJson: string,
         attachments: BacktraceAttachment[],
+        abort?: AbortSignal,
     ): Promise<BacktraceReportSubmissionResult<BacktraceSubmissionResponse>>;
 
     /**
      * Post data to Backtrace API
      * @param submissionUrl data submission URL
      * @param payload request payload
```
konraddysput, line 23: we need to add abort signal definition to the function description

packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
```diff
@@ -84,12 +89,11 @@ export class BacktraceMetrics implements BacktraceModule {
...
      * Sends event to the metrics system.
      */
-    public send() {
+    public async send(abortSignal?: AbortSignal) {
         if (!this._sessionProvider.shouldSend()) {
             return false;
         }
-        this.sendUniqueEvent();
-        this._summedEventsSubmissionQueue.send();
+        await Promise.all([this.sendUniqueEvent(abortSignal), this._summedEventsSubmissionQueue.send(abortSignal)]);
```
konraddysput, line 96: we do not need to wait on this tbh. We can just fire and forget that. Await can block API, so I think it would be better to not use it

konraddysput (changes requested):

packages/sdk-core/src/common/DelayHelper.ts
perf2711: as long as it builds, it's fine. And it builds 😄

packages/sdk-core/src/common/DelayHelper.ts
perf2711: Yes, you have to handle `delay` exceptions in case of aborting. I do this in `BacktraceMetrics`.

packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
perf2711: We cannot, because aborting will throw an unhandled promise rejection.

packages/sdk-core/src/common/DelayHelper.ts
konraddysput: How about my proposal? I believes it fits better to our code. Wdyt?

packages/sdk-core/src/model/http/BacktraceRequestHandler.ts
perf2711: Done.

packages/sdk-core/src/common/DelayHelper.ts
perf2711: It makes no difference at all, but changed.

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -268,6 +268,7 @@ export abstract class BacktraceCoreClient<O extends BacktraceConfiguration = Bac
         error: Error | string,
         attributes?: Record<string, unknown>,
         attachments?: BacktraceAttachment[],
+        abortSignal?: AbortSignal,
```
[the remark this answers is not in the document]
perf2711: Added.

PR #188, opened 2023-10-27
electron: add additional attributes to main crashes

perf2711:
This PR adds attributes from:
* available `BrowserWindow`s,
* `screen` - available displays,
* `net` - `isOnline`.

Jira task: BT-775

### DIFF
--- a/packages/electron/src/main/BacktraceClient.ts
+++ b/packages/electron/src/main/BacktraceClient.ts
@@ -1,8 +1,10 @@
 import {
-    BacktraceClient as NodeBacktraceClient,
     BacktraceNodeClientSetup,
     BacktraceSetupConfiguration,
+    BacktraceClient as NodeBacktraceClient,
 } from '@backtrace/node';
+import { BrowserWindow } from 'electron';
+import { WindowAttributeProvider, WindowAttributeProviderOptions } from './attributes/WindowAttributeProvider';
 import { BacktraceClientBuilder } from './builder/BacktraceClientBuilder';
 import { BacktraceMainElectronModule } from './modules/BacktraceMainElectronModule';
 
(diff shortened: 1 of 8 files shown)


packages/electron/src/main/BacktraceClient.ts
```diff
@@ -37,4 +47,35 @@ export class BacktraceClient extends NodeBacktraceClient {
...
+     * @param options Options for the attribute provider.
+     *
+     * @example
+     * // Will add attributes for window under `electron.window.${myWindow.id}`
+     * client.addWindowAttributes(myWindow);
+     *
+     * // Will add attributes for window under `electron.window.myWindow`
+     * client.addWindowAttributes(myWindow, { key: 'myWindow' })
+     */
+    public addWindow(window: BrowserWindow, options?: WindowAttributeProviderOptions): this {
```
konraddysput, line 63: is this useful to have? I believe it would be hard to query anything in the query builder based on the window-specific order. I believe it's worth including it in annotations however I'm not sure about attributes.

konraddysput (commented):
Can we verify my comments first? I'm ok with accepting the pull request, however, I'm not sure if we need all windows attributes as windows attributes rather I believe we need to have them in the annotations tab.

packages/electron/src/main/BacktraceClient.ts
perf2711: That's why it's optional, and you can specify the `key`, so the attributes will start with `window.{key}.*`.

examples/sdk/electron/src/main/index.ts
```diff
@@ -94,8 +94,15 @@ const createWindow = () => {
         },
     });
 
+    client.addMainWindow(win);
+
     win.loadFile('assets/index.html');
     win.webContents.openDevTools({ mode: 'detach' });
+
+    setTimeout(() => {
```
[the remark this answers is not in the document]
perf2711: whoops, didn't want this to be there, removed

konraddysput (approved):

PR #194, opened 2023-11-06
Hermes react stack trace

konraddysput:
# Why

This diff adds better support for Hermes-specific stack trace - it's like v8 but before source code information there might be "address at" instead of "at" like in typical v8 stack trace.

### DIFF
--- a/packages/react-native/src/BacktraceClient.ts
+++ b/packages/react-native/src/BacktraceClient.ts
@@ -42,7 +42,7 @@ export class BacktraceClient extends BacktraceCoreClient<BacktraceConfiguration>
             },
             requestHandler: new BacktraceBrowserRequestHandler(clientSetup.options),
             debugIdMapProvider: new VariableDebugIdMapProvider(global as DebugIdContainer),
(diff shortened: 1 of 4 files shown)


packages/react-native/src/BacktraceClient.ts
```diff
@@ -42,7 +42,7 @@ export class BacktraceClient extends BacktraceCoreClient<BacktraceConfiguration>
             },
             requestHandler: new BacktraceBrowserRequestHandler(clientSetup.options),
             debugIdMapProvider: new VariableDebugIdMapProvider(global as DebugIdContainer),
-            stackTraceConverter: new ReactStackTraceConverter(new V8StackTraceConverter()),
+            stackTraceConverter: new ReactStackTraceConverter(new V8StackTraceConverter('address-at')),
```
perf2711, line 45: should it be `address at`?

konraddysput: yeah, it should be `address at`. Updated

perf2711 (approved):

PR #195, opened 2023-11-08
node: fix AlternatingFileWriter crash on high load

perf2711:


packages/node/src/BacktraceClient.ts
```diff
@@ -40,6 +46,7 @@ export class BacktraceClient extends BacktraceCoreClient<BacktraceConfiguration>
             breadcrumbsManager.setStorage(
                 FileBreadcrumbsStorage.create(
                     this.sessionFiles,
+                    fileSystem,
```
konraddysput, line 49: it should be taken from clientSetup if possible
```suggestion
                    clientSetup?.fileSystem ?? fileSystem,
```

packages/sdk-core/tests/_mocks/fileSystem.ts
```diff
@@ -2,11 +2,11 @@ import path from 'path';
 import { BacktraceFileAttachment } from '../../src/model/attachment';
 import { FileSystem } from '../../src/modules/storage/FileSystem';
 
-type MockedFileSystem = {
-    [K in keyof FileSystem]: jest.Mock<ReturnType<FileSystem[K]>, Parameters<FileSystem[K]>>;
+export type MockedFileSystem<T extends FileSystem> = {
```
konraddysput, line 5: can we fix the warning and add es-lint ignore?

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -27,12 +28,16 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
     constructor(
         private readonly _mainFile: string,
         private readonly _fallbackFile: string,
+        fileSystem: NodeFileSystem,
         maximumBreadcrumbs: number,
     ) {
-        this._writer = new AlternatingFileWriter(_mainFile, _fallbackFile, maximumBreadcrumbs);
+        this._writer = new AlternatingFileWriter(_mainFile, _fallbackFile, maximumBreadcrumbs, fileSystem);
```
konraddysput, line 34: It is not part of this pull request, however, can we adjust the code also here? The number of breadcrumbs shouldn't extend the desired number of breadcrumbs declared by the user
```suggestion
        this._writer = new AlternatingFileWriter(
            _mainFile,
            _fallbackFile,
            Math.floor(maximumBreadcrumbs / 2),
            _fileSystem,
        );
```

konraddysput (approved):
Consider adjusting our AlternatingFileWriter algorithms between node and react-native. I think we should see a performance improvement if we batch entries into one file write operation

packages/node/src/BacktraceClient.ts
perf2711: Fixed.

packages/sdk-core/tests/_mocks/fileSystem.ts
perf2711: Added.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: Changed.

PR #205, opened 2023-11-22
browser: add middleware modes to redux middleware, move JSON action to attributes

perf2711:
Adds following modes to the Redux middleware:
* `all` - include everything by default,
* `omit-values` - add the breadcrumb, but omit the values,
* `off` - disable the middleware.

When `omit-values` is specified, only `type` is added to attributes.

### DIFF
--- a/packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
+++ b/packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
@@ -1,7 +1,10 @@
 import { BacktraceBreadcrumbs } from '@backtrace/sdk-core/src';
 import { Action, configureStore, createSlice, Middleware, PayloadAction } from '@reduxjs/toolkit';
 import { BacktraceClient } from '../../src/BacktraceClient';
-import { createBacktraceReduxMiddleware } from '../../src/redux/BacktraceReduxMiddleware';
+import {
+    BacktraceReduxMiddlewareOptions,
+    createBacktraceReduxMiddleware,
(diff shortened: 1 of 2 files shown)


packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
```diff
@@ -149,20 +178,59 @@ describe('createBacktraceReduxMiddleware', () => {
...
+                const expected = getExpectedBreadcrumb(interceptedAction, 'all');
+                store.dispatch(addToTestArray('Message to add'));
+                expect(breadcrumbsSpy).toHaveBeenCalledWith(...expected);
+            });
+
+            it('Should add only action type as JSON to breadcrumb attributes when mode=omit-values', () => {
+                const interceptedAction = { type: 'expected-type', payload: { abc: '123' } };
+                const store = getStore({ interceptAction: () => interceptedAction, mode: 'omit-values' });
+                const breadcrumbsSpy = getBreadcrumbsSpy('info');
+                const expected = getExpectedBreadcrumb({ type: interceptedAction.type }, 'omit-values');
```
adamcronin42, line 222: shouldn't you pass the whole interceptedAction here?

adamcronin42 (approved):

packages/browser/tests/redux/backtraceReduxMiddlewareTests.spec.ts
perf2711: updated

PR #208, opened 2023-11-27
react-native: managed breadcrumb support

konraddysput:
# Why

This diff adds managed breadcrumbs support to the react-native library. Basic breadcrumbs are populated.

Side note:
- we want to avoid using react library inside react-native - builder includes our react/browser/sdk-core library in the bundler output. We want to reduce the size of the library since it matters a lot on mobile

### DIFF
--- a/packages/react-native/src/breadcrumbs/events/AppStateBreadcrumbSubscriber.ts
+++ b/packages/react-native/src/breadcrumbs/events/AppStateBreadcrumbSubscriber.ts
@@ -1,0 +1,63 @@
+import {
+    BreadcrumbLogLevel,
+    BreadcrumbType,
+    type BacktraceBreadcrumbs,
+    type BreadcrumbsEventSubscriber,
+} from '@backtrace/sdk-core';
+import { AppState, Platform, type NativeEventSubscription } from 'react-native';
+
+export class AppStateBreadcrumbSubscriber implements BreadcrumbsEventSubscriber {
+    private _nativeEventSubscriptions: NativeEventSubscription[] = [];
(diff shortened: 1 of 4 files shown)


packages/react-native/src/breadcrumbs/events/AppStateBreadcrumbSubscriber.ts
```diff
@@ -0,0 +1,63 @@
...
+    BreadcrumbType,
+    type BacktraceBreadcrumbs,
+    type BreadcrumbsEventSubscriber,
+} from '@backtrace/sdk-core';
+import { AppState, Platform, type NativeEventSubscription } from 'react-native';
+
+export class AppStateBreadcrumbSubscriber implements BreadcrumbsEventSubscriber {
+    private _nativeEventSubscriptions: NativeEventSubscription[] = [];
+    public start(backtraceBreadcrumbs: BacktraceBreadcrumbs): void {
+        if ((backtraceBreadcrumbs.breadcrumbsType & BreadcrumbType.System) === BreadcrumbType.System) {
```
perf2711, line 12: why should the subscriber care if the level is `System` or not?

konraddysput: because we don't even need to set the connection between native and managed layer. Thats why I wrapped them in the if statement

perf2711: makes sense. Thanks

perf2711 (approved):

PR #220, opened 2023-12-08
cli: add-sources error behavior

perf2711:
This PR adds some changes to add-sources:
* by default, it will now warn about sources that failed to add if the source is not there, and skip it if the source is already in the sourcemap,
  * you can control what happens with failed sources using `--source-error-behavior`
* if `--force` is passed, all sources are treated as if they were not there

Jira task: BT-2080

### DIFF
--- a/tools/sourcemap-tools/src/SourceProcessor.ts
+++ b/tools/sourcemap-tools/src/SourceProcessor.ts
@@ -18,6 +18,25 @@ export interface ProcessResultWithPaths extends ProcessResult {
     readonly sourceMapPath: string;
 }
 
+export interface AddSourcesResult {
+    readonly sourceMap: RawSourceMap;
+
+    /**
+     * Source paths that were successfully added.
+     */
+    readonly succeeded: string[];
+
+    /**
+     * Source paths that failed to read, but source content was already in the sourcemap.
+     */
+    readonly skipped: string[];
+
+    /**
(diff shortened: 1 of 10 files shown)


tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -236,19 +256,35 @@ export class SourceProcessor {
             ? path.resolve(path.dirname(sourceMapPath), sourceMap.sourceRoot)
             : path.resolve(path.dirname(sourceMapPath));
 
-        const sourcesContent: string[] = [];
-        for (const sourcePath of sourceMap.sources) {
+        const succeeded: string[] = [];
+        const skipped: string[] = [];
+        const failed: string[] = [];
+
+        const sourcesContent: string[] = sourceMap.sourcesContent ?? [];
```
konraddysput, line 263: just checking:
if we skip the source content, because someone added a js file without type definition, will sources array be bigger than sourcesContent array?

konraddysput (approved):

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: no, I don't push to this array, I set the indices directly. If the file does not have an original source, the respective source content will be `null`.

PR #229, opened 2024-03-05
BT-718 - Reduce breadcrumb size - React Native

perf2711:
Follow up to https://github.com/backtrace-labs/backtrace-javascript/pull/228 for React Native.


konraddysput (commented):
Can we update branch?

PR #230, opened 2024-03-05
BT-718 - Reduce breadcrumb size - Node

perf2711:
Follow up to https://github.com/backtrace-labs/backtrace-javascript/pull/228 for Node.

### DIFF
--- a/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
+++ b/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
@@ -38,7 +38,7 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
             _fileSystem,
             _mainFile,
             _fallbackFile,
-            Math.floor((this._limits.maximumBreadcrumbs ?? 100) / 2),
+            this._limits.maximumBreadcrumbs ? Math.floor(this._limits.maximumBreadcrumbs / 2) : undefined,
             this._limits.maximumBreadcrumbsSize,
         );
     }
--- a/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
+++ b/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
@@ -3,6 +3,8 @@ import {
     Breadcrumb,
     BreadcrumbLogLevel,
     BreadcrumbsStorage,
+    BreadcrumbsStorageFactory,
+    BreadcrumbsStorageLimits,
     BreadcrumbType,
     jsonEscaper,
     RawBreadcrumb,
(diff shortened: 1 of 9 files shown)


packages/node/src/streams/lengthChunkSplitter.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param wholeLines If `true`, will split chunks before newlines, so whole lines are passed to the chunk.
+ */
+export function lengthChunkSplitter(maxLength: number, wholeLines: 'skip' | 'break' | false = false): ChunkSplitter {
+    let seen = 0;
+
+    const emptyBuffer = Buffer.of();
+
+    return function lengthChunkSplitter(data) {
+        const remainingLength = maxLength - seen;
+        if (data.length <= remainingLength) {
```
konraddysput, line 15: its weird that later we respect wholeLines information but when the data.length <= remainingLenght we don't. Actually why?

packages/node/src/streams/lengthChunkSplitter.ts
```diff
@@ -0,0 +1,51 @@
+import { ChunkSplitter } from './chunkifier.js';
+
+/**
+ * Splits data into chunks with maximum length.
+ * @param maxLength Maximum length of one chunk.
+ * @param wholeLines If `true`, will split chunks before newlines, so whole lines are passed to the chunk.
```
konraddysput, line 6: can we add a description for skip and break?

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -88,8 +114,15 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
         };
 
         const breadcrumbJson = JSON.stringify(breadcrumb, jsonEscaper());
-        this._writer.writeLine(breadcrumbJson);
+        const jsonLength = breadcrumbJson.length + 1; // newline
+        const sizeLimit = this._limits.maximumTotalBreadcrumbsSize;
+        if (sizeLimit !== undefined) {
+            if (jsonLength > sizeLimit) {
+                return id;
+            }
```
konraddysput, line 122: In the past id meant the id of the added breadcrumb. The breadcrumb id is no in the breadcrumb file. Should we return undefined instead? Or something that indicates that the breadcrumb was not added?

packages/node/src/streams/lengthChunkSplitter.ts
perf2711: because if all data fits into `remainingLength`, we don't need to split it. Ergo, line splitting does not occur.

konraddysput (approved):

PR #233, opened 2024-03-15
sdk-core: Return send result

konraddysput:
# Why

this diff adds a send result to the send method


packages/sdk-core/src/model/data/BacktraceSubmissionResult.ts
```diff
@@ -18,8 +18,16 @@ export class BacktraceReportSubmissionResult<T> {
         this._result = statusOrResponse;
     }
 
-    public static OnLimitReached<T>(): BacktraceReportSubmissionResult<T> {
-        return new BacktraceReportSubmissionResult('Limit reached', 'Client report limit reached');
+    public static OnLimitReached<T>(target: 'Server' | 'Client' = 'Server'): BacktraceReportSubmissionResult<T> {
+        return new BacktraceReportSubmissionResult('Limit reached', `${target} report limit reached`);
+    }
+
+    public static DisabledSdk<T>(): BacktraceReportSubmissionResult<T> {
```
perf2711, line 25: `SdkDisabled` sounds a bit better to me

perf2711 (approved):

packages/sdk-core/src/model/data/BacktraceSubmissionResult.ts
konraddysput: good idea!

PR #239, opened 2024-04-04
session-replay: add session replay package

perf2711:
This PR adds the session replay package, which enables us to attach session replays to reports sent to Backtrace.

See JSDocs and browser example for usage exampels.

### DIFF
--- a/packages/session-replay/package.json
+++ b/packages/session-replay/package.json
@@ -1,0 +1,34 @@
+{
+    "name": "@backtrace/session-replay",
+    "version": "0.0.1",
+    "description": "",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "build": "tsc",
+        "clean": "tsc -b --clean && rimraf \"lib\"",
+        "format": "prettier --write '**/*.ts'",
+        "lint": "eslint . --ext .ts",
+        "watch": "tsc -w",
+        "test": "NODE_ENV=test jest"
+    },
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/backtrace-labs/backtrace-javascript.git"
+    },
+    "author": "Backtrace <team@backtrace.io>",
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/backtrace-labs/backtrace-javascript/issues"
+    },
+    "files": [
+        "/lib"
+    ],
+    "homepage": "https://github.com/backtrace-labs/backtrace-javascript#readme",
+    "peerDependencies": {
+        "@backtrace/sdk-core": "^0.2.0"
+    },
+    "dependencies": {
+        "rrweb": "^2.0.0-alpha.4"
+    }
+}
--- a/packages/session-replay/package.json
+++ b/packages/session-replay/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@backtrace/session-replay",
     "version": "0.0.1",
-    "description": "",
+    "description": "Backtrace-JavaScript Session Replay module",
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
     "scripts": {
--- a/packages/session-replay/package.json
+++ b/packages/session-replay/package.json
@@ -9,8 +9,7 @@
         "clean": "tsc -b --clean && rimraf \"lib\"",
         "format": "prettier --write '**/*.ts'",
         "lint": "eslint . --ext .ts",
-        "watch": "tsc -w",
-        "test": "NODE_ENV=test jest"
+        "watch": "tsc -w"
     },
     "repository": {
         "type": "git",
--- a/packages/session-replay/src/BacktraceSessionRecorder.ts
+++ b/packages/session-replay/src/BacktraceSessionRecorder.ts
@@ -53,12 +53,7 @@ export class BacktraceSessionRecorder implements BacktraceAttachment {
     }
 
     public get(): string {
-        let events = [...(this._events ?? []), ...(this._previousEvents ?? [])];
-        if (this._options.maxTime) {
-            const cutout = Date.now() - this._options.maxTime;
-            events = events.filter((e) => e.timestamp >= cutout);
-        }
-
+        const events = [...(this._events ?? []), ...(this._previousEvents ?? [])];
         return JSON.stringify(events);
     }
 
--- a/packages/session-replay/src/BacktraceSessionRecorder.ts
+++ b/packages/session-replay/src/BacktraceSessionRecorder.ts
@@ -1,204 +1,7 @@
 import { BacktraceAttachment } from '@backtrace/sdk-core';
-import { MouseInteractions, eventWithTime } from '@rrweb/types';
+import { eventWithTime } from '@rrweb/types';
 import { record } from 'rrweb';
-import { MaskInputFn, MaskInputOptions, MaskTextFn } from 'rrweb-snapshot';
-import { recordOptions } from 'rrweb/typings/types';
-
-interface BacktraceSessionRecorderSamplingOptions {
-    /**
-     * Controls whether mouse movement is recorded.
-     * @default true
-     */
-    readonly mousemove?: boolean | number;
-
-    /**
-     * Controls specific mouse interactions.
-     *
-     * Can be one of:
-     * * `true` - all mouse interactions will be recorded
-     * * `false` - no mouse interactions will be recorded
-     * * `{[string]: boolean}` - control specific mouse interactions:
-     *     * `MouseUp`
-     *     * `MouseDown`
-     *     * `Click`
-     *     * `ContextMenu`
-     *     * `DblClick`
-     *     * `Focus`
-     *     * `Blur`
-     *     * `TouchStart`
-     *     * `TouchMove_Departed`
-     *     * `TouchEnd`
-     *     * `TouchCancel`
-     * @default true
-     * @example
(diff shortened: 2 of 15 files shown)


packages/session-replay/package.json
```diff
@@ -0,0 +1,34 @@
+{
+    "name": "@backtrace/session-replay",
+    "version": "0.0.1",
+    "description": "",
```
konraddysput, line 4: We need to update package.json

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
+import { BacktraceAttachment } from '@backtrace/sdk-core';
+import { MouseInteractions, eventWithTime } from '@rrweb/types';
+import { record } from 'rrweb';
+import { MaskInputFn, MaskInputOptions, MaskTextFn } from 'rrweb-snapshot';
+import { recordOptions } from 'rrweb/typings/types';
+
+interface BacktraceSessionRecorderSamplingOptions {
```
konraddysput, line 7: can we move options to another file? It takes a lot of space

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+    private readonly _maxTime?: number;
+
+    private _previousEvents?: eventWithTime[] = [];
+    private _events?: eventWithTime[] = [];
+
+    private _stop?: () => void;
+
+    constructor(private readonly _options: BacktraceSessionRecorderOptions) {
+        this._events = [];
+        this._maxEventCount = defaultIfNotFalse(_options.maxEventCount, 100);
```
konraddysput, line 224: isn't 
`_options.maxEventCount ?? 100`
enough?

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+        this._maxTime = defaultIfNotFalse(_options.maxTime, undefined);
+    }
+
+    public name = 'bt-session-replay-0';
```
konraddysput, line 228: can we move variables before constructor?

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    public start() {
+        const stop = record({
```
konraddysput, line 231:
```suggestion
        this._stop = record({
```

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+                ...this._options.advancedOptions,
+            },
+            emit: (event, isCheckout) => {
+                if (isCheckout || !this._events) {
+                    this._previousEvents = this._events;
+                    this._events = [];
+                }
+
+                this._events.push(event);
+                console.log(event);
```
konraddysput, line 248:
```suggestion

```

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+                scroll: this._options.sampling?.scroll,
+                ...this._options.advancedOptions,
+            },
+            emit: (event, isCheckout) => {
+                if (isCheckout || !this._events) {
+                    this._previousEvents = this._events;
+                    this._events = [];
+                }
+
+                this._events.push(event);
```
konraddysput, line 247: what in the situation when someone wants to remove specific element from the event? I think we need to give a user an option to modify the event.

konraddysput (changes requested):

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,276 @@
...
+                console.log(event);
+
+                if (this._options.advancedOptions?.emit) {
+                    this._options.advancedOptions.emit(event as never, isCheckout);
```
konraddysput, line 251: why its better to use never than unknown in this case? Or use a type of event from the library

packages/session-replay/src/BacktraceSessionRecorder.ts
perf2711: No, we want to give the user the option to disable this limit (e.g. instead of it, just use the time limit option).

packages/session-replay/src/BacktraceSessionRecorder.ts
perf2711: Changed.

packages/session-replay/src/BacktraceSessionRecorder.ts
perf2711: They should use the privacy options. Nevertheless, I added an `inspect` function to options.

konraddysput (approved):

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        if (this._options.privacy?.inspect) {
+            const inspected = this._options.privacy.inspect(event);
+            if (!inspected) {
+                return;
+            }
+            event = inspected;
+        }
+
+        this._events.push(event);
```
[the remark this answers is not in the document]
perf2711: This is managed by `isCheckout` param, and this is supplied to us by `rrweb` based on `checkoutEveryNth`/`checkoutEveryNms` options. When `onEmit` is called with `isCheckout == true`, old events are cleared.

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,80 @@
...
+    private readonly _maxTime?: number;
+
+    private _previousEvents?: eventWithTime[] = [];
+    private _events?: eventWithTime[] = [];
+
+    private _stop?: () => void;
+
+    constructor(private readonly _options: BacktraceSessionRecorderOptions) {
+        this._events = [];
+        this._maxEventCount = defaultIfNotFalse(_options.maxEventCount, 100);
```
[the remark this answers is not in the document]
perf2711: The user can disable a limit and keep it from setting to default. For example, if they want to use only `maxTime`, they would want to set `maxEventCount` to `false` to prevent it being set by default to 100.

packages/session-replay/src/BacktraceSessionRecorder.ts
```diff
@@ -0,0 +1,80 @@
...
+            sampling: {
+                mousemove: this._options.sampling?.mousemove,
+                mouseInteraction: this._options.sampling?.mouseInteraction,
+                input: this._options.sampling?.input,
+                media: this._options.sampling?.media,
+                scroll: this._options.sampling?.scroll,
+                ...this._options.advancedOptions,
+            },
+            emit: (event, isCheckout) => this.onEmit(event, isCheckout),
+            checkoutEveryNth: this._maxEventCount && Math.ceil(this._maxEventCount / 2),
```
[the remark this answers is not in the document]
perf2711: Each event group has to be "checked out" by `rrweb`, i.e. made sure that it can function on it's own. If you have a maximum of 100 events, the checkout will be every 50 events to make sure that we do sent at least 50 events after reset and not 0. If we would set it to 100, we would be sending a maximum of 200 events if we didn't want to clear the first group. Not sure if I explained this clearly 😅

packages/session-replay/src/BacktraceSessionRecorder.ts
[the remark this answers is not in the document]
adamcronin42: Makes sense. Makes me wonder if it might be a confusing pattern for the user. Would it make more sense to just have a separate field for disabling maxEventCount and maxTime that would default to being enabled and take precedence if disabled over any max set?

packages/session-replay/src/BacktraceSessionRecorder.ts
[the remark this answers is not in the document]
adamcronin42: So it's to give essentially two chances (50 events, 50 events) for a max of 100 in case the checkout of the first 50 fails, for example?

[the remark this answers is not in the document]
perf2711: It's goes more like this:
1. add first event, checkout - initialize empty `_events` array and add the event to it
2. add 49 events (so we have 50 events)
3. add 51st event, checkout - move `_events` to `_previousEvents`, reinitialize empty `_events` array and add the event to it
4. add 49 events (so we have 100 events)
5. add 101st event, checkout - move `_events` to `_previousEvents` (that drops the previous `_previousEvents`), reinitialize empty `_events` array and add the event to it
6. and so on...

This ensures that after adding 50 events, 50-100 events will be always available, and we do not exceed the 100 events limit.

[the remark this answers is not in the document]
adamcronin42: Ahh, got it. Thanks!

packages/session-replay/src/BacktraceSessionRecorder.ts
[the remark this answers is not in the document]
konraddysput: Agree with Adam, it's confusing. As the user I wouldn't expect maxEventCount can accept false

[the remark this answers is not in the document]
perf2711: I added new options to disable the limits.

PR #244, opened 2024-04-26
iOS source map support

konraddysput:
# Why

This diff simplifies Backtrace source map support for iOS in react-native.

ref: BT-761


rick-bt (approved):

examples/sdk/reactNative/README.md
```diff
@@ -19,3 +24,38 @@ android-sourcemap.sh script.
...
+# if you dont use hermes support, please skip this step.
+export HERMES_CLI_PATH="$(pwd)/../ios-hermesc.sh"
+
+/bin/sh -c "$WITH_ENVIRONMENT $REACT_NATIVE_XCODE"
+
+# copy javascript build output to the build directory
+cp "$CONFIGURATION_BUILD_DIR/main.jsbundle" $SOURCE_MAP_DIR
+
+# process source map with javascript code
+backtrace-js run --config "$(pwd)/../.backtracejsrc" --path "$SOURCE_MAP_DIR/main.jsbundle"
```
perf2711, line 56: maybe we should use `npx` here? 

```
npx --yes backtrace-js run --config "$(pwd)/../.backtracejsrc" --path "$SOURCE_MAP_DIR/main.jsbundle"
```

(the `--yes` switch will install `backtrace-js` if it is not available).

examples/sdk/reactNative/README.md
```diff
@@ -10,6 +10,11 @@ This example app shows features available in the @backtrace/react-native package
 
 #### Source maps
 
+Before executing any step:
+
+-   Please update .backtracejsrc file with your symbols submission URL and your sourcemap settings.
+-   please install @backtrace/javascript-cli library.
```
perf2711, line 16: AFAIK the user would have to install this package globally for line 56 to work. Could we use `npx` instead?

konraddysput: Good idea!

examples/sdk/reactNative/README.md
konraddysput: Good idea!

perf2711 (approved):

PR #245, opened 2024-05-06
Exit on disabled database

konraddysput:
# Why

We want to exit fast when there is no reason to process reports. This diff allows to exit from the database send method when the database is disabled (for example: via dispose method)

### DIFF
--- a/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
+++ b/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
@@ -203,6 +203,9 @@ export class BacktraceDatabase implements BacktraceModule {
         await this.send(abortSignal);
         const records = this.get().filter((n) => n.timestamp <= start);
         for (const record of records) {
+            if (abortSignal?.aborted) {
+                return;
+            }
             this.remove(record);
         }
     }
--- a/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
+++ b/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
@@ -18,6 +18,10 @@ export class BacktraceDatabase implements BacktraceModule {
         return this._enabled;
     }
 
+    /**
+     * Abort controller to cancel asynchronous database operations when
+     * the database is being disabled by the user.
+     */
     private readonly _abortController = new AbortController();
 
     private readonly _databaseRecordContext: BacktraceDatabaseContext;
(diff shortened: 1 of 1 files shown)


packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -206,12 +213,19 @@ export class BacktraceDatabase implements BacktraceModule {
             const records = [...this._databaseRecordContext.getBucket(bucketIndex)];
 
             for (const record of records) {
+                if (!this.enabled) {
+                    return;
+                }
```
perf2711, line 218: can you change this to [throwIfAborted](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted)? or, return some kind of information that the sending was aborted?

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -206,12 +213,19 @@ export class BacktraceDatabase implements BacktraceModule {
...
                 if (record.locked) {
                     continue;
                 }
                 try {
                     record.locked = true;
-                    const result = await this._requestHandler.send(record.data, record.attachments);
+                    const result = await this._requestHandler.send(
```
perf2711, line 224: I think we should also allow passing `AbortSignal` to the `send` method.

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
konraddysput: good idea. Thx!

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
konraddysput: good idea. thx

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -188,30 +196,37 @@ export class BacktraceDatabase implements BacktraceModule {
...
      */
-    public async flush() {
+    public async flush(abortSignal?: AbortSignal) {
         const start = TimeHelper.now();
-        await this.send();
+        await this.send(abortSignal);
         const records = this.get().filter((n) => n.timestamp <= start);
         for (const record of records) {
             this.remove(record);
         }
```
perf2711, line 207: should we remove only records that were actually sent, if `flush` was aborted?

konraddysput: No, we shouldn't. The expectation is to clear the database via flush. Send will also take care of cleaning them if we successfully send them.

konraddysput: if someone would use abort signal in this context the expectation would be to stop the execution. I think you're correct. I added an if statement to the loop.

perf2711 (approved):

PR #247, opened 2024-05-08
Release 0.3.2

konraddysput:
# Why

This pull request updates all our libraries versions

### DIFF
--- a/packages/browser/CHANGELOG.md
+++ b/packages/browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Version 0.3.1
 
--   added a new header to `BacktraceNodeRequestHandler`
+-   added a new HTTP header to report submission layer
 
 # Version 0.3.0
 
--- a/packages/browser/CHANGELOG.md
+++ b/packages/browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Version 0.3.1
 
--   added a new HTTP header to report submission layer
+-   added a new HTTP header to report submission layer (#246)
 
 # Version 0.3.0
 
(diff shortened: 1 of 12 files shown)


packages/node/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.3.1
+
+-   added a new HTTP header to report submission layer
```
perf2711, line 3: can you add here the merge request link?

packages/react/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.3.1
+
+-   added a new HTTP header to report submission layer
```
perf2711, line 3: can you add here the merge request link?

packages/browser/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.3.1
```
perf2711, line 1: Maybe it would be a good idea to sync this version to sdk-core version?

packages/node/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.3.1
```
perf2711, line 1: Maybe it would be a good idea to sync this version to sdk-core version?

perf2711 (approved):

packages/browser/CHANGELOG.md
konraddysput: I disagree, there will be a situation when a specific SDK will be updated without sdk-core update. I prefer to leave it as it is and continue updating libraries this way.

packages/node/CHANGELOG.md
konraddysput: I disagree, there will be a situation when a specific SDK will be updated without sdk-core update. I prefer to leave it as it is and continue updating libraries this way.

PR #249, opened 2024-06-17
Include windows in CICD pipeline. Use @v4

konraddysput:
# Why

This diff updates github acitons to @v4 and include windows cicd pipeline.


perf2711:
I think the pattern in Prettier should be changed to "*.ts".

perf2711 (approved):

PR #257, opened 2024-06-21
session-replay: PII defaults

perf2711:
This PR adds the following:
* pass privacy options to `rrweb` settings,
* mask all inputs by default,
* add input to browser example to demonstrate masking.

### DIFF
--- a/examples/sdk/browser/index.html
+++ b/examples/sdk/browser/index.html
@@ -24,6 +24,7 @@
                 <h1 class="card-header" style="text-align: center">Welcome to the Backtrace demo</h1>
                 <p style="text-align: center">Please pick one of the available options:</p>
                 <br />
+                <input class="center" type="text" placeholder="Test Session Replay Input!">
                 <div class="action-container center">
                     <div class="action-button center">
                         <a class="text" id="send-error" target="_blank"> Send an error</a>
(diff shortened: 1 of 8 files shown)


konraddysput (approved):

adamcronin42 (commented):
Very cool! I believe this only masks the text in input elements by default. I think we want to mask ALL text by default and then allow the user to selectively remove masking however they'd like. A way I was able to mask all text by default is via a `maskTextFn()` in `rrweb.record`:
`maskTextFn(t: string, el: HTMLElement | null) {
    	// if, for example, the user adds `data-unmask-example` we will not mask the text
        if (el?.matches('[data-unmask-example]')) {
            return t;
        } else {
            // mask all text
            return t.replace(/[\S]/g, '*');
        }
    }`

This is just an example, but we could make it work for different selectors, classes, etc or even let the user overwrite this maskTextFn.

hmm the formatting is ugly, I'll just ping it to you.

perf2711:
@adamcronin42 I added some changes to text masking:
* added options: 
  * `maskAllText`
  * `unmaskTextClass`
  * `unmaskTextSelector`
* set `maskTextSelector` to `*` in `rrweb` options to have `rrweb` call `maskTextFn` with all elements,
* our `maskTextFn` takes into account options for masking/unmasking
  * if `maskAllText` is `true`, you have to explicitly unmask elements and have them not matched by mask selectors
  * if `maskAllText` is `false`, you have to explicitly mask elements and have them not matched by unmask selectors

PR #260, opened 2024-07-12
Add support for error.type for electron crashes

konraddysput:
# Why

This diff adds support for the error.type attribute for electron crashes. 

ref: BT-3057

### DIFF
--- a/packages/electron/src/main/modules/BacktraceMainElectronModule.ts
+++ b/packages/electron/src/main/modules/BacktraceMainElectronModule.ts
@@ -1,10 +1,11 @@
 import {
+    AttributeType,
     BacktraceData,
     BacktraceModule,
     BacktraceModuleBindData,
     RawBreadcrumb,
     SubmissionUrlInformation,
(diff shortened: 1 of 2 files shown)


packages/electron/src/main/modules/BacktraceMainElectronModule.ts
```diff
@@ -96,14 +97,19 @@ export class BacktraceMainElectronModule implements BacktraceModule {
                 app.setPath('crashDumps', options.database.path);
             }
 
+            const getElectronDefaultAttributes = (attributes: Record<string, AttributeType> = attributeManager.get('scoped').attributes) => {
```
perf2711, line 100: instead of supplying a default value here, why not just pass it on line 108?

konraddysput: I think its a better idea. I'm chaning my code.

perf2711 (approved):

PR #262, opened 2024-07-26
electron, node: upload attachments after native crashes

perf2711:
## SDK Core

Changes to database have been made:
* database accepts now also attachment records,
* `scoped` and `dynamic` attachments are now available (attachment data is still resolved when reports are sent, but this is required for `FileAttachmentsManager` in Node to know which attachments should be stored,
* `getAttachmentProviders` now is used when available in `BreadcrumbsStorage`,
* `sendAttachment` and `postAttachment` methods have been added to `BacktraceReportSubmission` and `BacktraceRequestHandler`, respectively,
* several APIs have been exposed to child packages.

## Node

Node SDK will now add attachments from previous sessions, if they are available.
* `FileBreadcrumbsStorage`: `createFromSession` has been replaced with `getSessionAttachments` ,
* `FileAttachmentsManager` has been added - this tracks scoped attachments in a file in database,
* `postAttachment` has been implemented in `NodeBacktraceRequestHandler`,
* file system APIs have been exposed to child packages.

## Electron

Crashpad in Electron does not support attaching files.
We circumvent that by sending attachments to reports created by Crashpad.
* `sendAttachment` has been implemented in `IpcReportSubmission`
* attachments from reports created in Crashpad are added to database


packages/sdk-core/src/modules/database/model/BacktraceDatabaseRecord.ts
```diff
@@ -1,15 +1,34 @@
...
 import { BacktraceData } from '../../../model/data/BacktraceData';
 
-export interface BacktraceDatabaseRecord {
+export interface ReportBacktraceDatabaseRecord {
+    readonly type: 'report';
     readonly data: BacktraceData;
     readonly id: string;
     readonly hash: string;
     readonly timestamp: number;
+    readonly sessionId?: string;
```
konraddysput, line 10: how can we benefit from sessionId?

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -68,6 +62,19 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
         ];
     }
 
+    public getAttachmentProviders(): BacktraceAttachmentProvider[] {
+        return [
+            {
+                get: () => new BacktraceFileAttachment(this._mainFile, 'bt-breadcrumbs-0'),
```
konraddysput, line 68: I'm not fully sure what it means that the attachment is dynamic in this context. I was sure the attachment is dynamic when the attachment context is in the app memory. breadcrumbs are not in the memory. Why it's dynamic?

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -39,20 +40,13 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
         );
     }
 
-    public static createFromSession(
-        session: SessionFiles,
-        fileSystem: NodeFileSystem,
-    ): FileBreadcrumbsStorage | undefined {
+    public static getSessionAttachments(session: SessionFiles) {
```
konraddysput, line 43: I think we're breaking api here. Why we cannot continue using fileSystem?

konraddysput (commented):
The biggest problem I see in this pull request that for me is a show stopper, is the problem related to the database behavior. 

As the user, if I set number of records in the database to X, I expect, the database can store 8 records. In this situation, we can drop some reports for attachments. In practice, those attachments are not reports and in the end the user prefers to see reports and most likely is fine with loosing attachments. 


If someone has also more attachments than the database size, we will have some problems with that.

perf2711:
You're right - I forgot about database limits being filled by attachments.

To address this, I've changed how the database counts records. Now they will be limited per type, i.e. you can have 5 reports, and 100 attachments.

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
```diff
@@ -1,53 +1,145 @@
...
 import { FileSystem } from '../storage';
-import { BacktraceDatabaseRecord } from './model/BacktraceDatabaseRecord';
+import {
+    AttachmentBacktraceDatabaseRecord,
+    BacktraceDatabaseRecord,
+    ReportBacktraceDatabaseRecord,
+} from './model/BacktraceDatabaseRecord';
 
-export class BacktraceDatabaseFileRecord implements BacktraceDatabaseRecord {
+export const BacktraceDatabaseFileRecord = {
```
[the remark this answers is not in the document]
perf2711: image

I cannot mark this as class, as this breaks the type below.

Not sure what you mean by a logger? What do you want to log?

packages/sdk-core/src/modules/database/model/BacktraceDatabaseRecord.ts
perf2711: We can lock specific sessions and not all of them. This was used before, but was hidden behind `application.session` attribute.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: The files may not exist at a given time. Also, with the changes made in breadcrumb limiting, this will not work, as file names are incremented instead of switched.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: It's just not required anymore, as we're not creating `FileBreadcrumbsStorage`. Instead we just return attachments.

packages/sdk-core/src/modules/database/BacktraceDatabaseFileRecord.ts
```diff
@@ -1,53 +1,145 @@
...
+            return new ReportBacktraceDatabaseFileRecord(record, attachments);
         } catch {
             return undefined;
         }
     }
+}
 
-    private static isFileAttachment(attachment: BacktraceAttachment): attachment is BacktraceFileAttachment {
-        return 'filePath' in attachment && typeof attachment.filePath === 'string';
+export class AttachmentBacktraceDatabaseFileRecord implements AttachmentBacktraceDatabaseRecord {
```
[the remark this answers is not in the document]
perf2711: done

packages/sdk-core/src/modules/attachments/AttachmentManager.ts
```diff
@@ -0,0 +1,62 @@
+import { Events } from '../../common/Events';
+import { AttachmentEvents } from '../../events/AttachmentEvents';
+import { BacktraceAttachment } from '../../model/attachment';
+import { BacktraceAttachmentProvider } from './BacktraceAttachmentProvider';
+
+export class AttachmentManager {
```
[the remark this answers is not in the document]
perf2711: done

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -41,7 +46,10 @@ export class BacktraceDatabase implements BacktraceModule {
         private readonly _sessionFiles?: SessionFiles,
     ) {
         this._databaseRecordContext = new BacktraceDatabaseContext(this._options?.maximumRetries);
-        this._maximumRecords = this._options?.maximumNumberOfRecords ?? 8;
+        this._recordLimits = {
+            report: this._options?.maximumNumberOfRecords ?? 8,
+            attachment: this._options?.maximumNumberOfAttachmentRecords ?? 100,
```
konraddysput, line 51: one more idea - 100 is a random number and based on this we might have some problems. Maybe we should just agree on 5 attachments per session - 30? Usually, users want to send 1-2 attachments + 1-2 for breadcrumbs, so 5/session is a safe value

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -247,25 +313,45 @@ export class BacktraceDatabase implements BacktraceModule {
      * Prepare database to insert records
      * @param totalNumberOfRecords number of records to insert
```
konraddysput, line 314: param needs to be updated

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -115,23 +125,63 @@ export class BacktraceDatabase implements BacktraceModule {
...
-        this.prepareDatabase();
+        const sessionId = backtraceData.attributes?.['application.session'];
 
-        const record: BacktraceDatabaseRecord = {
-            count: 1,
+        const record: ReportBacktraceDatabaseRecord = {
+            type: 'report',
             data: backtraceData,
             timestamp: TimeHelper.now(),
             hash: '',
```
konraddysput, line 139: i think we can also drop hash support if we're dropping count - by using hash we could count the same errors. Just in case, this doesn't need to happen in this pull request - this is information for the future

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -115,23 +125,63 @@ export class BacktraceDatabase implements BacktraceModule {
...
+
+        const record: AttachmentBacktraceDatabaseRecord = {
+            type: 'attachment',
+            timestamp: TimeHelper.now(),
+            hash: '',
+            id: IdGenerator.uuid(),
+            rxid,
+            locked: false,
+            attachment,
+            sessionId: typeof sessionId === 'string' ? sessionId : undefined,
```
konraddysput, line 181: can we add attachment without sessionid?

konraddysput (commented):
I need to verify one more thing in the code and most likely I will be ready to accept pull request. Thanks!

packages/electron/src/main/modules/BacktraceMainElectronModule.ts
```diff
@@ -121,6 +141,64 @@ export class BacktraceMainElectronModule implements BacktraceModule {
...
+
+                const fileAttachments = FileAttachmentsManager.createFromSession(session, fileSystem);
+                const attachments = await fileAttachments.get();
+
+                for (const attachment of attachments) {
+                    database.addAttachment(rxid, attachment, session.sessionId);
+                }
+
+                // Write an empty crash lock, so we know that this crash is already taken care of
+                await fileSystem.writeFile(crashLock, '');
```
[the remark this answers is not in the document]
perf2711: yes, session cleanup removes everything matching its session ID

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
perf2711: 30 is as much arbitrary as 100 is. Default maximum number of previous sessions is 1, though this could be changed, and not necessarily available. I'll change this to 10, as it seems more reasonable. This doesn't cost us disk space, as each attachment record is little in size.

packages/electron/src/main/modules/BacktraceMainElectronModule.ts
```diff
@@ -121,6 +141,64 @@ export class BacktraceMainElectronModule implements BacktraceModule {
...
+
+                const crashLock = session.getFileName(`electron-crash-lock-${rxid}`);
+                // If crash lock exists, do not attempt to add attachments twice
+                if (await fileSystem.exists(crashLock)) {
+                    continue;
+                }
+
+                for (const attachment of FileBreadcrumbsStorage.getSessionAttachments(session)) {
+                    database.addAttachment(rxid, attachment, session.sessionId);
+                }
```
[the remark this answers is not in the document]
perf2711: Updated.

packages/electron/src/main/modules/BacktraceMainElectronModule.ts
```diff
@@ -121,6 +141,64 @@ export class BacktraceMainElectronModule implements BacktraceModule {
...
+        }
+
+        await database.send();
+    }
+
+    private getCrashRxid(crashId: string): string | undefined {
+        try {
+            return JSON.parse(crashId)._rxid;
+        } catch {
+            return crashId.match(/"_rxid":"([a-f0-9-]+)"/i)?.[1];
```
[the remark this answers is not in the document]
perf2711: I changed the regular expression to be more strict.

packages/sdk-core/src/model/configuration/BacktraceDatabaseConfiguration.ts
```diff
@@ -26,6 +26,13 @@ export interface EnabledBacktraceDatabaseConfiguration {
      */
     maximumNumberOfRecords?: number;
 
+    /**
+     * The maximum number of attachments stored in the offline database. When the limit is reached,
```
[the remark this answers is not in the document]
perf2711: There was a default value before. I changed it to use `@default`.

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
perf2711: Done

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
perf2711: I guess not. I made it mandatory.

konraddysput (approved):

PR #265, opened 2024-08-19
sdk-core, browser, react-native: fix abort event not being removed from signal

perf2711:
Every time `anySignal` is used, an `abort` event is added to the passed `AbortSignal`.

If the signal is not cancelled, the event is not removed. If the code using `anySignal` is executed multiple times, this can cause events to build up without being cleaned up.

This PR changes this and introduces a function `disposeSignal` that cleans up the signal afterwards.

### DIFF
--- a/packages/browser/src/BacktraceBrowserRequestHandler.ts
+++ b/packages/browser/src/BacktraceBrowserRequestHandler.ts
@@ -44,13 +44,14 @@ export class BacktraceBrowserRequestHandler implements BacktraceRequestHandler {
     ): Promise<BacktraceReportSubmissionResult<T>> {
         const controller = new AbortController();
         const id = setTimeout(() => controller.abort(), this._timeout);
+        const signal = anySignal(abortSignal, controller.signal);
 
         try {
             const response = await fetch(submissionUrl, {
                 method: 'POST',
                 body: payload,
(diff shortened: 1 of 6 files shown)


packages/browser/src/BacktraceBrowserRequestHandler.ts
```diff
@@ -44,13 +44,14 @@ export class BacktraceBrowserRequestHandler implements BacktraceRequestHandler {
     ): Promise<BacktraceReportSubmissionResult<T>> {
         const controller = new AbortController();
         const id = setTimeout(() => controller.abort(), this._timeout);
+        const [signal, disposeSignal] = anySignal(abortSignal, controller.signal);
```
konraddysput, line 47: I just notice - if someone pas a custom abort signal to the request handler, the setTimeout function will not work, right?

konraddysput (approved):

packages/browser/src/BacktraceBrowserRequestHandler.ts
perf2711: why not? that's the whole reason for `anySignal`, to abort when any signal aborts

konraddysput: you're correct - i forgot what anysignal does.

perf2711:
note - `sdk-core` should be released as minor version update, as this is a breaking change vs other packages that may be using this.

PR #266, opened 2024-08-22
sdk-core: use ESM modules

perf2711:
This PR adds ESM module support to `sdk-core`.

`sdk-core` is now bundled into two files:
- `bundle.mjs` - ESM modules

### DIFF
--- a/packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
+++ b/packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
@@ -25,8 +25,8 @@ export class BacktraceMetrics implements BacktraceModule {
     public readonly DEFAULT_UPDATE_INTERVAL = TimeHelper.convertSecondsToMilliseconds(30 * 60);
     public readonly DEFAULT_SERVER_URL = 'https://events.backtrace.io';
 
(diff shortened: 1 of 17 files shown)


packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
```diff
@@ -38,6 +38,9 @@ export class BacktraceMetrics implements BacktraceModule {
         private readonly _summedEventsSubmissionQueue: MetricsQueue<SummedEvent>,
         private readonly _uniqueEventsSubmissionQueue: MetricsQueue<UniqueEvent>,
     ) {
+        this.metricsHost = this._options.metricsSubmissionUrl ?? this.DEFAULT_SERVER_URL;
```
konraddysput, line 41: do we need this change?

konraddysput (approved):

packages/sdk-core/src/modules/metrics/BacktraceMetrics.ts
perf2711: yes, it errors out, presumably because we use `this._options` here

konraddysput (approved):

PR #267, opened 2024-08-22
node: use ESM modules

perf2711:
This PR adds ESM module support to `node`.

`node` is now bundled into two files:
- `bundle.mjs` - ESM modules


packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
```diff
@@ -57,19 +58,20 @@ export class ApplicationInformationAttributeProvider implements BacktraceAttribu
...
-        if (require.main?.path) {
+        if (typeof require !== 'undefined' && require.main?.path) {
             possibleSourcePaths.unshift(path.dirname(require.main.path));
         }
         return possibleSourcePaths;
     }
 
     private generatePathBasedOnTheDirName() {
-        const nodeModulesIndex = __dirname.lastIndexOf('node_modules');
+        const dirname = fileURLToPath(path.dirname(import.meta.url));
```
konraddysput, line 68: are you sure it will work with package/console?

konraddysput (approved):

packages/node/src/attributes/ApplicationInformationAttributeProvider.ts
perf2711: should be fine. This is what is generated in cjs bundle:
```
const dirname = url.fileURLToPath(path.dirname((typeof document === 'undefined' ? require('u' + 'rl').pathToFileURL(__filename).href : (_documentCurrentScript && _documentCurrentScript.src || new URL('bundle.cjs', document.baseURI).href))));   
```

PR #270, opened 2024-08-22
electron: use ESM modules

perf2711:
This PR adds ESM module support to `electron`.

`electron` is now bundled into files:
- `main.mjs`, `renderer.mjs`, `preload.mjs` - ESM modules 
- `main.cjs`, `renderer.cjs`, `preload.cjs` - CommomJS (Node) modules


konraddysput:
Electron doesn't build here

packages/electron/package.json
```diff
@@ -39,5 +93,14 @@
     },
     "peerDependencies": {
         "electron": "12 - 28"
```
konraddysput, line 95: side note outside this review:
should we bumb electron version to 32?

konraddysput (approved):

packages/electron/package.json
perf2711: sure

PR #271, opened 2024-08-22
nestjs: use ESM modules

perf2711:
This PR adds ESM module support to `nestjs`.

`nestjs` is now bundled into two files:
- `bundle.mjs` - ESM modules


packages/nestjs/package.json
```diff
@@ -41,23 +43,23 @@
         "/lib"
     ],
     "devDependencies": {
+        "@backtrace/node": "^0.3.1",
```
konraddysput, line 46: why backtrace/node is a dev dependency here?

konraddysput (approved):

packages/nestjs/package.json
perf2711: good catch, should be in normal deps

PR #272, opened 2024-08-22
session-replay: use ESM modules

perf2711:
This PR adds ESM module support to `session-replay`.

`session-replay` is now bundled into two files:
- `bundle.mjs` - ESM modules

This change also fixes import errors in `react-scripts`.


konraddysput (changes requested):
Session replay package doesnt build here

perf2711:
@konraddysput it builds now.

konraddysput (approved):

PR #277, opened 2024-08-30
ci: smoke tests

perf2711:
This PR adds smoke tests support.

Currently, Node is only tested.


smoketests/node/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "smoketest-node",
+    "version": "1.0.0",
+    "description": "@backtrace/node smoketests",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "scripts": {
+        "start": "node lib/index.js",
+        "build": "tsc",
+        "watch": "npm run build -- --watch",
+        "trust-ssl-cert": "cp /certs/cert.crt /usr/local/share/ca-certificates/ && update-ca-certificates"
```
konraddysput, line 11: do we need to have it here?

smoketests/node-esm/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "version": "1.0.0",
+    "description": "@backtrace/node smoketests",
+    "main": "lib/index.js",
+    "types": "lib/index.d.ts",
+    "type": "module",
+    "scripts": {
+        "start": "node lib/index.js",
+        "build": "tsc",
+        "watch": "npm run build -- --watch",
+        "trust-ssl-cert": "cp /certs/cert.crt /usr/local/share/ca-certificates/ && update-ca-certificates"
```
konraddysput, line 12: do we need to have it here?

smoketests/node-esm/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+});
+
+console.log('submitting message');
+client.send('test message').then((result) => {
+    if (result.status !== 'Ok') {
+        console.error('Unexpected result:', result);
+        process.exit(1);
+    }
+
+    console.log(result.result?._rxid);
```
konraddysput, line 35: my recommendation - we should also check if the rxId is defined.

konraddysput (approved):

smoketests/node-esm/src/index.ts
perf2711: this is checked by the smoketests themselves:
```
expect(result.stdout.trim()).toMatch(RXID_REGEX);
```

smoketests/node/package.json
perf2711: nope, forgot to remove it, removed

smoketests/node-esm/package.json
perf2711: nope, forgot to remove it, removed

PR #283, opened 2024-09-06
ci: additional smoke tests

perf2711:
This PR adds new smoke tests for Node, testing combinations of JS/TS and ESM/CJS.

This also adds a WDIO smoke test run on Sauce Labs for React SDK with Session Replay.


.github/workflows/test.yml
```diff
@@ -50,33 +77,34 @@ jobs:
...
               env:
                   SMOKETESTS_SUBMIT_LAYER_URL: ${{ secrets.SMOKETESTS_SUBMIT_LAYER_URL }}
                   SMOKETESTS_DIRECT_SUBMIT_URL: ${{ secrets.SMOKETESTS_DIRECT_SUBMIT_URL }}
 
     test_windows:
         runs-on: windows-latest
+
+        strategy:
+            matrix:
+                node-version: [18.x]
```
konraddysput, line 96: do we need this matrix if we test only against 18?

package.json
```diff
@@ -13,6 +13,8 @@
         "rebuild": "npm run clean && npm run build",
         "test": "npm run test --ws --if-present",
         "smoketest": "npm run start --workspace=smoketests",
+        "smoketest:node": "npm run start:jest --workspace=smoketests",
```
konraddysput, line 16: idea: maybe we should have just one command called smoke test and execute it for all packages that have it in package.json - the same way how we run test or build

smoketests/packages/node-js-cjs/index.cjs
```diff
@@ -0,0 +1,38 @@
...
+    database: {
+        enable: true,
+        path: path.join(process.cwd(), 'database'),
+        captureNativeCrashes: true,
+        createDatabaseDirectory: true,
+    },
+});
+
+console.log('submitting message');
+client.send('test message').then((result) => {
```
konraddysput, line 30: can we please send more? at least an exception + message. We can also send unhandled exception if we instruct node correctly.

smoketests/packages/react-ts-esm/src/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+            }),
+        ),
+);
+
+function App() {
+    const [status, setStatus] = useState('loading');
+    const [rxid, setRxid] = useState('');
+    const [error, setError] = useState('');
+
+    useEffect(() => {
```
konraddysput, line 37: idea: if we can use buttons - we can click some of them and trigger actions instead of just veryfing what is the page content. Right now it's impossible to test against an error. Testign if the "message" is sent, is kinda useless.

package.json
perf2711: It cannot be this way, because we don't want to run browser smoketests on each workflow on GitHub.

.github/workflows/test.yml
perf2711: I think it's plain and simple to see then on which Node versions it is tested.

smoketests/packages/react-ts-esm/src/index.tsx
perf2711: Added some new tests.

smoketests/packages/node-js-cjs/index.cjs
perf2711: Added some new tests.

smoketests/tests/__helpers/npm.ts
```diff
@@ -1,17 +1,50 @@
-import { SpawnOptionsWithoutStdio } from 'child_process';
-import { asyncSpawn } from './asyncSpawn';
+import fs from 'fs/promises';
+import path from 'path';
+import { asyncSpawn, AsyncSpawnOptions, AsyncSpawnResult } from './asyncSpawn.js';
 
-export function npm(cwd: string, args: string[], options?: SpawnOptionsWithoutStdio) {
+export function npm(cwd: string, args: string[], options?: AsyncSpawnOptions) {
```
konraddysput, line 5: we talked about this on Slack but I still dont get it:
Questions:
- Why do we need to write scripts around this. Why we cannot just use CICD pipeline setup.
- if we're already in the npm workspace, why the development process is hard? We can just have a watch process. We don't need to install every time npm packages or clean them. 
- Why it's not enough to use npm workspace and the same actions. We can execute all actions in a single workspace that has a task node:e2e or browser:e2e. 
- Why do we need to have a dedicated project for smoke tests?

konraddysput (commented):
e2e tests are important and I'm happy that we're doing this. I'm not fully sure what is the idea for the smoketest project - how it should work in the dev environment or in cicd pipeline. I have a lot of open questions related to flow. Does it make sense to also add a README file with information what is this project and what it does/how to use it?


I'm happy to accept the pull request once we confirm

smoketests/tests/prepare.ts
```diff
@@ -0,0 +1,35 @@
...
+let packagePaths = process.argv.slice(2).map((p) => path.join(PACKAGE_DIR, p));
+if (!packagePaths.length) {
+    packagePaths = await discoverPackages();
+}
+
+for (const packagePath of packagePaths) {
+    console.log(`======= PACKAGE ${path.relative('.', packagePath)} =======`);
+    console.log();
+    console.log(`[${path.relative('.', packagePath)}] cleaning up package`);
+    await cleanPackage(packagePath);
```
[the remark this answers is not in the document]
perf2711: Removed mandatory cleaning.

smoketests/packages/node-js-cjs/index.cjs
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { BacktraceClient } = require('@backtrace/node');
+const path = require('path');
+
+const SUBMISSION_URL = process.argv[2];
```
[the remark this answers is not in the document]
perf2711: yes, it's passed in jest tests.

smoketests/tests/prepare.ts
```diff
@@ -0,0 +1,35 @@
+import fs from 'fs/promises';
```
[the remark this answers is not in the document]
perf2711: I want the scripts to be available to the developer as well.

smoketests/wdio.conf.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    user: process.env.SMOKETESTS_SAUCE_USERNAME,
+    key: process.env.SMOKETESTS_SAUCE_ACCESS_KEY,
+
+    region: 'eu',
+    specs: ['./tests/wdio/**/*.spec.ts'],
+    exclude: [],
+    maxInstances: 10,
+    capabilities: [
+        {
```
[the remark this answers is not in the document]
perf2711: I added iOS. I had some trouble executing it on Android, so I skipped this for now.

smoketests/tests/__helpers/npm.ts
perf2711: >Why do we need to write scripts around this. Why we cannot just use CICD pipeline setup.
We want the developer to be able to run the tests without CICD pipelines.

>if we're already in the npm workspace, why the development process is hard? We can just have a watch process. We don't need to install every time npm packages or clean them.
You can definitely do that if you want. Install NPM packages yourself, set up a watch process, and run the test.

>Why it's not enough to use npm workspace and the same actions. We can execute all actions in a single workspace that has a task node:e2e or browser:e2e.
Installing the packages by itself is a test. I wanted to separate this.

>Why do we need to have a dedicated project for smoke tests?
Why not? Where you would put it otherwise?

konraddysput (approved):

PR #290, opened 2024-09-10
sdk-core: use core client as event emitter, send reports from database via client

perf2711:
This PR adds event support for client directly.

### Client events
You can attach to `before-skip`, `before-send`, and `after-send` events on client instance:

```typescript
const client = BacktraceClient.initialize({ ... });
client.on('after-send', (report, data, attachments, result) => { ... });
```

This is required for me to use in smoketests, so I know if `after-send` is executed after e.g. an exception is thrown.

### Database events
`BacktraceDatabase` now emits events of it's own:

```ts
type BacktraceDatabaseEvents = {
    added: [record: BacktraceDatabaseRecord];
    removed: [record: BacktraceDatabaseRecord];
    'before-send': [record: BacktraceDatabaseRecord];
    'after-send': [
        record: BacktraceDatabaseRecord,
        result: BacktraceReportSubmissionResult<BacktraceSubmissionResponse>,
    ];
};
```

### DIFF
--- a/packages/sdk-core/src/BacktraceCoreClient.ts
+++ b/packages/sdk-core/src/BacktraceCoreClient.ts
@@ -1,6 +1,6 @@
 import { CoreClientSetup } from './builder/CoreClientSetup.js';
 import { Events } from './common/Events.js';
-import { ReportEvents } from './events/ReportEvents.js';
+import { ClientEvents } from './events/ClientEvents.js';
 import {
     BacktraceAttachment,
     BacktraceAttributeProvider,
@@ -34,7 +34,9 @@ import { MetricsBuilder } from './modules/metrics/MetricsBuilder.js';
 import { SingleSessionProvider } from './modules/metrics/SingleSessionProvider.js';
 import { RateLimitWatcher } from './modules/rateLimiter/RateLimitWatcher.js';
 
-export abstract class BacktraceCoreClient<O extends BacktraceConfiguration = BacktraceConfiguration> {
+export abstract class BacktraceCoreClient<
+    O extends BacktraceConfiguration = BacktraceConfiguration,
+> extends Events<ClientEvents> {
     /**
      * Backtrace client instance
      */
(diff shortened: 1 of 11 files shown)


perf2711:
> since this is the internal api, maybe we can drop the deprecated field once we update all SDKs?

It turned out that this was only used in `sdk-core`. I removed this altogether.

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -16,7 +18,17 @@ import {
     ReportBacktraceDatabaseRecord,
 } from './model/BacktraceDatabaseRecord.js';
 
-export class BacktraceDatabase implements BacktraceModule {
+type BacktraceDatabaseEvents = {
```
konraddysput, line 21: can we keep this class clean and move it to another file?

konraddysput (approved):

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -329,14 +326,18 @@ export abstract class BacktraceCoreClient<O extends BacktraceConfiguration = Bac
 
         const submissionAttachments = this.generateSubmissionAttachments(report, reportAttachments);
 
-        this.reportEvents.emit('before-send', report, backtraceData, submissionAttachments);
+        return this.sendData(backtraceData, submissionAttachments, abortSignal);
+    }
+
+    public async sendData(backtraceData: BacktraceData, attachments: BacktraceAttachment[], abortSignal?: AbortSignal) {
```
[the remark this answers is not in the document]
perf2711: yeah, this is removed already

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
perf2711: why? it's directly correlated with this class and this class only

konraddysput: Database configuration also. If we're making an excuse just for events, we will end up in the really bad state for this file.

perf2711: moved to another file

PR #293, opened 2024-10-02
nestjs: add exception filter and exception handler

perf2711:
This PR adds `BacktraceExceptionFilter` and `BacktraceExceptionHandler`.

The filter allows to catch more exceptions than previously possible with `BacktraceInterceptor`.
The handler allows to use the exception handling logic in own code.

### DIFF
--- a/packages/nestjs/README.md
+++ b/packages/nestjs/README.md
@@ -13,6 +13,8 @@ after which you can explore the rich set of Backtrace features.
     - [Integrate the SDK](#integrate-the-sdk)
     - [Upload source maps](#upload-source-maps)
     - [Add a Backtrace error interceptor](#add-a-backtrace-error-interceptor)
+    - [Add a Backtrace exception filter](#add-a-backtrace-exception-filter)
+    - [Use the Backtrace exception handler in your code](#use-the-backtrace-exception-handler-in-your-code)
 1. [Error Reporting Features](#error-reporting-features)
     - [Attributes](#attributes)
     - [File Attachments](#file-attachments)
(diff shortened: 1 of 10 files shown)


packages/nestjs/src/backtrace.handler.ts
```diff
@@ -0,0 +1,208 @@
...
+        };
+    }
+
+    private getWsAttributes(ws: WsArgumentsHost) {
+        return {
+            ['ws.data']: ws.getData(),
+        };
+    }
+
+    private filterException(exception: unknown, filter: ExceptionTypeFilter): boolean {
```
konraddysput, line 194: you're not filtering exception but rather you're checking if the exception match filter criteria. It was hard for me to check if you have a bug in the "sendError" method, because you used the same way the filterException method. 

Maybe a different name can improve readability? for example: `filterIncludeError` ?

konraddysput (approved):

packages/nestjs/README.md
```diff
@@ -75,11 +77,38 @@ class AppController {
...
+                excludeExceptionTypes: (error) => error instanceof HttpException && error.getStatus() < 500,
+            },
+        }),
+    ],
+    controllers: [AppController],
+})
+class AppModule {}
+```
+
+As shown in the example above, `includeExceptionTypes` and `excludeExceptionTypes` accept either an array of error
```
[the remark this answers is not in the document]
perf2711: good idea, added.

packages/nestjs/src/backtrace.handler.ts
perf2711: I renamed it to `testException` and added some docs.

PR #294, opened 2024-10-03
Conditional smoke test

konraddysput:
This pull request allows to conditionally run smoke tests on specific code change

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,16 +24,33 @@ jobs:
             - run: npm run lint
             - run: npm run format:check
             - run: npm test
+            - uses: dorny/paths-filter@v3
+                id: filter
+                with:
+                    filters: |
+                    browser:
+                        - '.github/packages/browser/**'
+                        - '.github/packages/sdk-core/**'
+                        - '.github/packages/sdk-core/**'
+                    node:
+                        - '.github/packages/sdk-core/**'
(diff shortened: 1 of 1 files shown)


.github/actions/smoke-tests/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  using: "composite"
+  steps:
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+          filters: |
+                browser:
+                    - 'packages/browser/**'
+                    - 'packages/sdk-core/**'
+                    - 'packages/sdk-core/**'
```
perf2711, line 28: `sdk-core` is specified twice. Can you add `react` instead?

.github/actions/smoke-tests/action.yml
```diff
@@ -0,0 +1,49 @@
...
+                node:
+                    - 'packages/sdk-core/**'
+                    - 'packages/electron/**'
+                    - 'packages/nestjs/**'
```
perf2711, line 32: I think there's no point in rerunning node smoketest if electron/nestjs changes.

konraddysput:
we need to add an option to ignore web/node tests for specific use cases so we can reuse this code in multiple other places (for example: we can avoid running e2e tests when there is no change in browser/node package)

perf2711 (approved):

PR #296, opened 2024-10-08
sdk-core, node: attribute validation

perf2711:
This PR adds attribute validation on initialize, which checks the following attributes:
* `application`
* `application.version`

This also removes the check from `node`.

### DIFF
--- a/packages/sdk-core/src/BacktraceCoreClient.ts
+++ b/packages/sdk-core/src/BacktraceCoreClient.ts
@@ -218,6 +218,8 @@ export abstract class BacktraceCoreClient<
     }
 
     public initialize() {
+        this.validateAttributes();
+
         if (this.fileSystem && this.options.database?.createDatabaseDirectory) {
             if (!this.options.database.path) {
                 throw new Error(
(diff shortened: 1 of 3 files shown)


packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -418,4 +420,51 @@ export abstract class BacktraceCoreClient<
...
+
+        // Validate scoped attributes first. If they return no errors, there is no point
+        // in checking all attributes, which resolving of may be slower.
+        const scoped = this.attributeManager.get('scoped');
+        const scopedErrors = validate(scoped.attributes);
+        if (!Object.keys(scopedErrors).length) {
+            return;
+        }
+
+        const allAttributes = this.attributeManager.get();
```
konraddysput, line 458:
```suggestion
        const dynamicAttributes = this.attributeManager.get('dynamic');
        const allAttributes = {...dynamicAttributes, ...scoped};
```

packages/sdk-core/src/BacktraceCoreClient.ts
```diff
@@ -418,4 +420,51 @@ export abstract class BacktraceCoreClient<
     private static destroy() {
         this._instance = undefined;
     }
+
+    private validateAttributes() {
+        function isNotEmptyString(v: unknown) {
```
konraddysput, line 425: we're introducing a nice way to validate attributes, but in our scenario, we don't need such complexity. What we can do is just simply test if those two attributes are defined. Based on what we want to verify and how easily we can achieve that, wouldn't it be better if we keep the smallest/simplest solution for now?

konraddysput (changes requested):
Something is wrong with our node tests. I think we expect the app to fail when there is no application version or application in the attributes.

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: how would you do this more "easily"?

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: It may not be so simple. `get` iterates over all attribute providers, and then tests them if they are scoped/dynamic. For instance, if we would have these providers:
```
const provider1 = { type: 'scoped': get: () => ({ a: 'foo' });
const provider2 = { type: 'dynamic': get: () => ({ a: '' });
const provider3 = { type: 'scoped': get: () => ({ a: 'bar' });
```

`allAttributes` would be `{a: ''}`, while `get()` would return `{ a: 'bar' }`.

konraddysput: good point!

packages/sdk-core/src/BacktraceCoreClient.ts
konraddysput: just simply 
```
const attributes = this.attributeManager.ger()
const validAttributes = !!attributes["application"] && !!attributes["application.version"]
```

konraddysput (approved):

packages/sdk-core/src/BacktraceCoreClient.ts
perf2711: simplified a bit

konraddysput (approved):

PR #301, opened 2024-10-15
react-native: android crash handler upgrade

konraddysput:
# Why

This pull request upgrades the native crash reporting solution to the latest version from backtrace-android. All changes introduced in the: https://github.com/backtrace-labs/backtrace-android/pull/128/files are now also available here. 

Why do we need those? 

The new version of the AGP prevents executables from being added to NDK libraries. Due to this, we cannot use the crash handler the same way. With these changes, instead of having an executable, that sends a crash code, we have a java class responsible for a minidump generation.

ref: BT-5028


packages/react-native/android/src/main/java/backtrace/library/BacktraceReactNative.java
```diff
@@ -49,29 +49,42 @@ public String getName() {
...
-                values,
-                attachmentPaths.toArrayList().toArray(new String[0]),
-                false,
-                null
-        );
-
+        BacktraceCrashHandlerWrapper nativeCommunication = new BacktraceCrashHandlerWrapper();
+
+        Boolean result =
+                new File(handlerPath).exists()
```
perf2711, line 66: can you add a comment that describes the difference between `initializeCrashHandler` and `initializeJavaCrashHandler`?

packages/react-native/android/src/main/java/backtrace/library/common/AbiHelper.java
```diff
@@ -0,0 +1,14 @@
+package backtraceio.library.common;
+
+import android.os.Build;
+
+public class AbiHelper {
+    public static String getCurrentAbi() {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+            // on newer Android versions, we'll return only the most important Abi version
+            return Build.SUPPORTED_ABIS[0];
```
perf2711, line 9: are we always sure that `SUPPORTED_ABIS` length will be at least 1?

packages/react-native/android/src/main/java/backtrace/library/models/nativeHandler/CrashHandlerConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+        // extend system-specific environment variables, with variables needed to properly run app_process via crashpad
+        File nativeLibraryDirectory = new File(nativeLibraryDirPath);
+
+        String backtraceNativeLibraryPath = getBacktraceNativeLibraryPath(nativeLibraryDirPath, apkPath, arch);
+        File allNativeLibrariesDirectory = nativeLibraryDirectory.getParentFile();
+        String allPossibleLibrarySearchPaths = TextUtils.join(File.pathSeparator, new String[]{
+                nativeLibraryDirPath,
+                allNativeLibrariesDirectory.getPath(),
+                System.getProperty("java.library.path"),
+                "/data/local"});
```
perf2711, line 59: won't this get joined as `...path//data/local`?

packages/react-native/android/src/main/java/backtrace/library/models/nativeHandler/CrashHandlerConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+        return environmentVariables;
+    }
+
+    public String useCrashpadDirectory(String databaseDirectory) {
+        String databasePath = databaseDirectory + CRASHPAD_DIRECTORY_PATH;
+        File crashHandlerDir = new File(databasePath);
+        // Create the crashpad directory if it doesn't exist
+        if (!crashHandlerDir.exists()) {
+            crashHandlerDir.mkdir();
+        }
```
perf2711, line 75:
```suggestion
        crashHandlerDir.mkdirs();
```

packages/react-native/android/src/main/java/backtrace/library/models/nativeHandler/CrashHandlerConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private String getBacktraceNativeLibraryPath(String nativeLibraryDirPath, String apkPath, String arch) {
+        String backtraceNativeLibraryPath = String.format("%s/%s", nativeLibraryDirPath, BACKTRACE_NATIVE_LIBRARY_NAME);
+        File backtraceNativeLibrary = new File(backtraceNativeLibraryPath);
+
+        // If ndk libraries are already extracted, we shouldn't use libraries from the apk.
+        // Otherwise. We need to find a path in the apk to use compressed libraries from there.
+        return backtraceNativeLibrary.exists()
+                ? backtraceNativeLibraryPath
+                : String.format("%s!/lib/%s/%s", apkPath, arch, BACKTRACE_NATIVE_LIBRARY_NAME);
```
perf2711, line 87: is the exclamation mark required here? if it is, can you add a comment why?

packages/react-native/src/crashReporter/CrashReporter.ts
```diff
@@ -1,4 +1,5 @@
 import { type AttributeType, type BacktraceAttachment, type FileSystem } from '@backtrace/sdk-core';
+import path from 'path';
```
perf2711, line 2: This causes build errors when I try to use the example.

perf2711 (changes requested):

packages/react-native/android/src/main/java/backtrace/library/BacktraceReactNative.java
konraddysput: done

packages/react-native/android/src/main/java/backtrace/library/models/nativeHandler/CrashHandlerConfiguration.java
konraddysput: no, it will be separated by `:`

output

```
/data/app/~~nlc7R6EdprMECkAbt0jRFw==/com.reactnative-u_yqZOmeMUdl3wdqYW_3_g==/lib/arm64:/data/app/~~nlc7R6EdprMECkAbt0jRFw==/com.reactnative-u_yqZOmeMUdl3wdqYW_3_g==/lib:/system/lib64:/system_ext/lib64:/data/local
```

perf2711 (approved):

PR #308, opened 2024-10-23
react-native: fix invalid RN object returned from iOS BacktraceReactNative.initialize function

perf2711:
This PR returns `nil` instead of `instance` from `BacktraceReactNative.initialize`, as the `instance` object was impossible to serialize to JSON.

https://reactnative.dev/docs/0.70/native-modules-ios#synchronous-methods
>The return type of this method must be of object type (id) and should be serializable to JSON. This means that the hook can only return nil or JSON values (e.g. NSNumber, NSString, NSArray, NSDictionary).

### DIFF
--- a/packages/react-native/ios/BacktraceReactNative.mm
+++ b/packages/react-native/ios/BacktraceReactNative.mm
@@ -15,7 +15,7 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(initialize:(NSString*)submissionUrl
     }
     instance = [[BacktraceCrashReporter alloc] initWithBacktraceUrl:submissionUrl andDatabasePath: databasePath andAttributes: attributes andOomSupport:TRUE andAttachments:attachmentPaths];
     [instance start];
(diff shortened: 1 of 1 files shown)


packages/react-native/ios/BacktraceReactNative.mm
```diff
@@ -15,7 +15,7 @@ @implementation BacktraceReactNative
     }
     instance = [[BacktraceCrashReporter alloc] initWithBacktraceUrl:submissionUrl andDatabasePath: databasePath andAttributes: attributes andOomSupport:TRUE andAttachments:attachmentPaths];
     [instance start];
-    return instance;
+    return nil;
```
konraddysput, line 18: do we need to return something from the function?

konraddysput (approved):

packages/react-native/ios/BacktraceReactNative.mm
perf2711: nope, the return value is not used at all.

PR #315, opened 2024-11-12
react-native: replace AlternatingFileWriter with WritableStream and ChunkifierSink for breadcrumbs

perf2711:
This PR replaces `AlternatingFileWriter` with `WritableStream` and `ChunkifierSink` based on `chunkifier` from Node.
As React Native does not have a `WritableStream` support by default, a ponyfill is used.


packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -21,32 +25,48 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
...
-            _fallbackFile,
-            Math.floor(maximumBreadcrumbs / 2),
-            _fileSystem,
+        this._sink = new FileChunkSink({
+            maxFiles: 2,
+            fs: this._fileSystem,
+            file: (n) => session.getFileName(FileBreadcrumbsStorage.getFileName(n)),
+        });
+
+        this._dest = new WritableStream(
```
konraddysput, line 43: what is dest and what is the difference between dest and writer? Maybe changing the variable name makes sense?

packages/react-native/src/storage/ReactNativeFileSystem.ts
```diff
@@ -89,4 +89,8 @@ export class ReactNativeFileSystem implements FileSystem {
     public createAttachment(path: string, name?: string | undefined): BacktraceAttachment<unknown> {
         return new BacktraceFileAttachment(this, path, name);
     }
+
+    public createWriteStream(path: string): FileWritableStream {
```
konraddysput, line 93: why do we want to provide a write stream for our users? I think we don't want to manage custom write stream issues our customers might ran into

packages/react-native/src/storage/StreamWriter.ts
```diff
@@ -17,3 +19,56 @@ export interface StreamWriter {
...
+        }
+    }
+
+    public async write(chunk: string) {
+        if (!this._streamId) {
+            throw new Error('File is not open.');
+        }
+
+        if (!(await this._streamWriter.append(this._streamId, chunk))) {
+            throw new Error(`Failed to write data to file ${this.path}.`);
```
konraddysput, line 63: what benefits bring throwing an exception here? Should we fail silently? In theory our library should never throw an error. Is there any consequence if we fail silently? If we fail loud, what the user can do to fix it?

packages/react-native/src/storage/Chunkifier.ts
```diff
@@ -0,0 +1,120 @@
...
+    private _context?: StreamContext<W>;
+    private _splitter?: ChunkSplitter<W>;
+    private _chunkCount = 0;
+
+    constructor(private readonly _options: ChunkifierOptions<W>) {}
+
+    public async write(data: W): Promise<void> {
+        // If data is empty from the start, forward the write directly to current stream
+        if (this.isEmpty(data)) {
+            return await this.ensureStreamContext().streamWriter.write(data);
```
konraddysput, line 51: do we need to run a write operation at all? Based on the stream writer implementation, we don't need to?

packages/react-native/src/storage/Chunkifier.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    private createStreamContext(): StreamContext<W> {
+        const stream = this._options.sink(this._chunkCount++);
+        const writer = stream.getWriter();
+        return { stream, streamWriter: writer, isEmptyChunk: true };
+    }
+
+    private reset() {
+        this._context = undefined;
+        this._splitter = undefined;
```
konraddysput, line 114: problem:

I needed to check multiple times splitter definition and implementation to understand every time when you reset the splitter you're changing the current line. Is it worth explaining it somewhere here?

packages/react-native/src/storage/FileChunkSink.ts
```diff
@@ -0,0 +1,83 @@
...
+            const stream = this.createStream(n);
+            this._streamTracker.push(stream);
+            return stream;
+        };
+    }
+
+    private createStream(n: number) {
+        const path = this._options.file(n);
+        // TODO: What to do if this returns undefined?
+        return this._options.fs.createWriteStream(path);
```
konraddysput, line 58: we can't perform any stream operation if it returns undefined.

packages/react-native/tests/storage/chunkifier.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        return [chunk.substring(0, index), chunk.substring(index + 1)];
+    };
+}
+
+function noopSplitter<W extends Chunk>(): ChunkSplitter<W> {
+    return (c: W) => [c];
+}
+
+describe('ChunkifierSink', () => {
+    it('should call splitter function with every chunk', async () => {
```
konraddysput, line 23: is it worth adding a case where splitter should not be called?

konraddysput (approved):

packages/react-native/src/storage/ReactNativeFileSystem.ts
perf2711: it's not for our users, it's for us

packages/react-native/src/storage/Chunkifier.ts
perf2711: It's just that any write to this sink should end up in writing something to the underlying stream.

packages/react-native/src/storage/FileChunkSink.ts
perf2711: yeah, this is outdated, forgot to remove this.

packages/react-native/src/storage/StreamWriter.ts
perf2711: We shouldn't fail silently here. However, I added `catch` to `write` in `FileBreadcrumbStorage` and `close` in `FileChunkSink`.

packages/react-native/src/storage/Chunkifier.ts
perf2711: I added some comments.

packages/react-native/tests/storage/chunkifier.spec.ts
perf2711: Added.

PR #316, opened 2024-11-13
browser, react, session-replay: add api-extractor (NFC)

perf2711:
This PR adds `@microsoft/api-extractor` to web packages to bundle declaration files.

For these packages, declaration files are built in `lib/types`, and from then, bundled to `lib/bundle.d.ts`. `lib/types` is NOT included in the NPM package.

This should fix TypeScript build errors in packages that use browser SDK.


perf2711:
Closing for now in favor of https://github.com/backtrace-labs/backtrace-javascript/pull/317.

PR #320, opened 2024-11-18
react-native: reduce breadcrumb size

perf2711:
Follow up to https://github.com/backtrace-labs/backtrace-javascript/pull/228 for React Native.

This uses the new breadcrumb writers implemented in https://github.com/backtrace-labs/backtrace-javascript/pull/315.

### DIFF
--- a/packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
+++ b/packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
@@ -8,12 +8,14 @@ import {
     type BacktraceAttachmentProvider,
     type Breadcrumb,
     type BreadcrumbsStorage,
+    type BreadcrumbsStorageFactory,
+    type BreadcrumbsStorageLimits,
     type RawBreadcrumb,
 } from '@backtrace/sdk-core';
 import { WritableStream } from 'web-streams-polyfill';
 import { BacktraceFileAttachment } from '..';
 import { type FileSystem } from '../storage';
-import { ChunkifierSink } from '../storage/Chunkifier';
+import { ChunkifierSink, type ChunkSplitterFactory } from '../storage/Chunkifier';
 import { FileChunkSink } from '../storage/FileChunkSink';
 import { lineChunkSplitter } from '../storage/lineChunkSplitter';
 
@@ -32,7 +34,7 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
     constructor(
         session: SessionFiles,
         private readonly _fileSystem: FileSystem,
-        maximumBreadcrumbs: number,
+        private readonly _limits: BreadcrumbsStorageLimits,
     ) {
         this._sink = new FileChunkSink({
             maxFiles: 2,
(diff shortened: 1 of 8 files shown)


packages/react-native/tests/storage/FileBreadcrumbsStorage.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+        const actualFallback = await loadBreadcrumbs(fs, fallbackStream);
+        expect(actualMain).toEqual(expectedMain);
+        expect(actualFallback).toEqual(expectedFallback);
+    });
+
+    it('should return no more than maximumBreadcrumbs breadcrumbs', async () => {
+        const fs = mockStreamFileSystem();
+        const session = new SessionFiles(fs, '.', 'sessionId');
+
+        const breadcrumbs: RawBreadcrumb[] = [
```
konraddysput, line 186: small nit: 
this file is quite long and can be simplified by moving breadcrumbs outside each test to a describe scope. You're always creating the same breadcrumbs so .. why not?

packages/react-native/tests/storage/FileBreadcrumbsStorage.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+                level: 'debug',
+                message: 'b',
+                timestamp: expect.any(Number),
+                type: 'http',
+            },
+        ];
+
+        const storage = new FileBreadcrumbsStorage(session, fs, {
+            maximumBreadcrumbs: 100,
+            maximumTotalBreadcrumbsSize: (JSON.stringify(expectedMain[0]).length + 10) * 2,
```
konraddysput, line 298: wouldn't it be better if you can sum lenght of expectedMain[0] + expectedMain[1] length instead of doing `*2` If you modify the test, it might be broken

packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -32,29 +35,48 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
...
 
-        this._destinationStream = new WritableStream(
-            new ChunkifierSink({
-                sink: this._sink.getSink(),
-                splitter: () => lineChunkSplitter(Math.ceil(maximumBreadcrumbs / 2)),
-            }),
-        );
+        const splitters: ChunkSplitterFactory<string>[] = [];
+        const maximumBreadcrumbs = this._limits.maximumBreadcrumbs;
+        if (maximumBreadcrumbs !== undefined) {
```
konraddysput, line 48: double checking: 
maximumBreadcrumbs should always be defined - otherwise, we can generate too big files and cause application problems. I can easily imagine a situation where the customer experiences the issue caused by us overflooding the SD card

konraddysput (approved):

packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
perf2711: It's up to the end user to define the valid options. By default, maximum breadcrumbs is 100, so when setting setting this to undefined, you know what you're doing.

You could also say that 10 breadcrumbs could fill up the SD card, if you put a billion bytes in each of them.

packages/react-native/tests/storage/FileBreadcrumbsStorage.spec.ts
perf2711: Why do we care about file length for tests? It's also easier to modify a test if the value is inside each case.

packages/react-native/tests/storage/FileBreadcrumbsStorage.spec.ts
perf2711: good catch, changed.

packages/react-native/src/breadcrumbs/FileBreadcrumbsStorage.ts
konraddysput: who wants that? in the end, we're the error-reporting solution, not a full logger. Our main goal is for our library never to crash the app. This is a potential problem.

perf2711: I don't know who wants that. Besides, you could always set the number to something like 2874523758979, and the effect would be the same. You could attach a million attributes to an error, and it could crash the app also.

PR #329, opened 2024-12-16
webpack-plugin: fix e2e tests (NFC)

perf2711:
Fixes errors in Webpack E2E tests:
* some mocks are not required anymore, so they are not mocked to simplify the test
* the tests failed due to some changes to the `sourcemap-tools` package, which changed the return types


tools/webpack-plugin/tests/e2e/createE2ETest.ts
```diff
@@ -19,9 +19,16 @@ export function createE2ETest(configBuilder: (mode: webpack.Configuration['mode'
         }
 
         function mockProcessor() {
-            return jest
-                .spyOn(SourceProcessor.prototype, 'processSourceAndSourceMapFiles')
-                .mockImplementation(async (_, __, debugId) => Ok({ debugId: debugId ?? 'debugId' } as never));
+            return jest.spyOn(SourceProcessor.prototype, 'processSourceAndSourceMapFiles');
+            // .mockImplementation(async (sourcePath, sourceMapPath, debugId) =>
```
konraddysput, line 23: 1. dead code - can we remove it?
2. Why are you changing it?

konraddysput (commented):
Can you share more details why you're removing/changing the test code? What doesn't work?

tools/webpack-plugin/tests/e2e/createE2ETest.ts
perf2711: 1. yes
2. there's no need to mock this

PR #332, opened 2025-01-21
sourcemap-tools: add processSource to SourceProcessor

konraddysput:
# Why

This change exposes an option in the sourcemap-tools to process source maps via translation process (for example: via metro in react-native) 

ref: BT-5175

### DIFF
--- a/tools/sourcemap-tools/src/SourceProcessor.ts
+++ b/tools/sourcemap-tools/src/SourceProcessor.ts
@@ -125,19 +125,19 @@ export class SourceProcessor {
      * @param force Force adding changes.
      * @returns Used debug ID, new source and new sourcemap.
      */
-    public async processSourceAndAvailableSourceMap(
+    private async processSourceAndAvailableSourceMap(
         source: string,
         sourceMap: RawSourceMap,
         debugId?: string,
         force?: boolean,
     ): Promise<ProcessResultWithSourceMaps>;
-    public async processSourceAndAvailableSourceMap(
+    private async processSourceAndAvailableSourceMap(
         source: string,
         sourceMap?: undefined,
         debugId?: string,
         force?: boolean,
     ): Promise<ProcessResultWithoutSourceMap>;
-    public async processSourceAndAvailableSourceMap(
+    private async processSourceAndAvailableSourceMap(
         source: string,
         sourceMap?: RawSourceMap,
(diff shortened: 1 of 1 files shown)


perf2711 (approved):
Could you try and extract some of the duplicated logic in `processSource` and `processSourceAndSourceMap` into functions?

tools/sourcemap-tools/src/SourceProcessor.ts
```diff
@@ -116,21 +145,26 @@ export class SourceProcessor {
...
         }
 
         if (force || !sourceDebugId || !this._debugIdGenerator.hasCommentSnippet(source, debugId)) {
             const sourceComment = this._debugIdGenerator.generateSourceComment(debugId);
             newSource = appendBeforeWhitespaces(newSource, '\n' + sourceComment);
         }
 
+        if (!sourceMap) {
+            return { debugId, source: newSource };
+        }
```
perf2711, line 167: I wonder why it doesn't show an error here? `ProcessResult` expects `sourceMap` to be set.

Can you add a new interface called `ProcessResultWithoutSourcemap` (this will reduce refactoring in other packages) and annotate this function like this:
```
    public async processSourceAndSourceMap(
        source: string,
        sourceMap: RawSourceMap,
        debugId?: string,
        force?: boolean,
    ): Promise<ProcessResult>;
    public async processSourceAndSourceMap(
        source: string,
        sourceMap?: undefined,
        debugId?: string,
        force?: boolean,
    ): Promise<ProcessResultWithoutSourcemap>;
    public async processSourceAndSourceMap(
        source: string,
        sourceMap?: RawSourceMap,
        debugId?: string,
        force?: boolean,
    ): Promise<ProcessResult> {
```

Also, change the return type on `processSource` to return `Promise<ProcessResultWithoutSourcemap>`.

perf2711 (changes requested):

tools/sourcemap-tools/src/SourceProcessor.ts
perf2711: So it DOES return an error!
image

perf2711: But not on build... strange.

perf2711 (approved):

PR #333, opened 2025-01-21
react-native: simplify source map processing flow

konraddysput:
# Why

This pull request simplifies the source map processing flow by adding source map support via metro. 

In addition to that, on:

* Android
There is a new gradle task that uploads a source map file in the end of the build/assemble process

* iOS
the xcode build phase is now simpler and requires less execution. 

Since we wanted to have simple and similar flows in both OSes, we wanted to continue with metro and adjust each build with new requirements.  

ref: BT-5175


examples/sdk/reactNative/android/app/build.gradle
```diff
@@ -47,7 +49,7 @@ react {
     // hermesCommand = "$rootDir/my-custom-hermesc/bin/hermesc"
     //
     //   The list of flags to pass to the Hermes compiler. By default is "-O", "-output-source-map"
-    // hermesFlags = ["-O", "-output-source-map"]
+    hermesFlags = ["-O", "-output-source-map"]
```
perf2711, line 52: if this is the default option, does this have to be uncommented?

examples/sdk/reactNative/metro.config.js
```diff
@@ -8,12 +9,16 @@ const path = require('path');
...
+    watchFolders: [
+        path.resolve('../../../packages/react-native'),
+        path.resolve('../../../packages/react-native/node_modules'),
+        path.resolve('../../../node_modules'),
+        path.resolve('../../../packages/sdk-core'),
+    ],
+    serializer: {
+        async customSerializer(entryPoint, preModules, graph, options) {
+            return backtraceSourceMapProcessor.processSourceMap(entryPoint, preModules, graph, options);
+        },
```
perf2711, line 21:
```suggestion
        customSerializer: backtraceSourceMapProcessor.processSourceMap
```

examples/sdk/reactNative/ios-sourcemap-upload.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ -z "$3" ]; then
+    echo "Error: Missing path to the .backtracejsrc file."
+    exit 1
+fi
+
+backtrace_configuration_path="$3"
+
+debug_id=$(<"$debug_id_file_path")
+
+jq ". += {\"debugId\": \"$debug_id\"}" "$source_map_file_path" > tmp.map && mv tmp.map $source_map_file_path
```
perf2711, line 47:
```suggestion
jq ". += {\"debugId\": \"$debug_id\"}" "$source_map_file_path" > "${source_map_file_path}.tmp" && mv "${source_map_file_path}.tmp" $source_map_file_path
```

examples/sdk/reactNative/ios-sourcemap-upload.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Check if the file exists
+if [ ! -f "$source_map_file_path" ]; then
+    echo "Error: File '$source_map_file_path' does not exist."
+    exit 1
+fi
+
+DEBUG_ID_PATH_ENV="DEBUG_ID_PATH"
+
+debug_id_file_path=${!DEBUG_ID_PATH_ENV:-${2:-}}
```
perf2711, line 31: what's the benefit of `DEBUG_ID_PATH_ENV`?

packages/react-native/android/upload-sourcemaps.gradle
```diff
@@ -0,0 +1,48 @@
...
+
+        def debugIdPathEnv = System.getenv("DEBUG_ID_PATH")
+
+        def debugIdFile = debugIdPathEnv ? file(debugIdPathEnv) : layout.buildDirectory
+                .dir("intermediates/sourcemaps/react/${buildVariant}/.backtrace-sourcemap-id")
+                .get()
+                .asFile
+
+        if (!debugIdFile.exists()) {
+            println("Backtrace: Cannot find .backtrace-sourcemap-id debug info.")
```
perf2711, line 22:
```suggestion
            println("Backtrace: Cannot find .backtrace-sourcemap-id file. Check if customSerializer has been set to Backtrace serializer in metro.config.js.")
```

packages/react-native/android/upload-sourcemaps.gradle
```diff
@@ -0,0 +1,48 @@
...
+            return
+        }
+
+        def sourcemapDestinationDirectory = layout.buildDirectory.dir("generated/sourcemaps/react").get().asFile
+        def mapFiles = fileTree(dir: sourcemapDestinationDirectory, include: '**/*.map').files
+        def mapFile = mapFiles ? mapFiles.iterator().next() : null
+
+        if (!mapFile) {
+            println("Backtrace: Cannot find final source map file.")
```
perf2711, line 31: could you elaborate which file do you mean? e.g. what was supposed to generate it?

packages/react-native/android/upload-sourcemaps.gradle
```diff
@@ -0,0 +1,48 @@
...
+            return
+        }
+
+        def jsonSlurper = new JsonSlurper()
+        def mapData = jsonSlurper.parse(mapFile)
+        mapData.debugId = debugIdFile.text
+        def jsonBuilder = new JsonBuilder(mapData)
+        mapFile.text = jsonBuilder.toPrettyString()
```
perf2711, line 39: not sure what was the previous serialization, but is pretty text required here? It will increase the map size

packages/react-native/android/upload-sourcemaps.gradle
```diff
@@ -0,0 +1,48 @@
...
+
+        // Execute Backtrace JS CLI to process upload processed source maps
+        def command = ["npx", "--yes", "@backtrace/javascript-cli", "upload", "-p", mapFile.absolutePath]
+        exec {
+            workingDir file("$rootProject.projectDir/..")
+            commandLine command
+        }
```
perf2711, line 46: is `@backtrace/javascript-cli` output shown in build output?

packages/react-native/processSourceMap.js
```diff
@@ -0,0 +1,62 @@
...
+
+    // no source map option set
+    const sourceMapOutputPathParameter = process.argv.indexOf('--sourcemap-output');
+    if (sourceMapOutputPathParameter === -1) {
+        return bundle;
+    }
+
+    const sourceMapOutputPath = process.argv[sourceMapOutputPathParameter + 1];
+    const backtraceSourceMapId =
+        DEBUG_ID_PATH ?? path.join(path.dirname(sourceMapOutputPath), '.backtrace-sourcemap-id');
```
perf2711, line 29: would it be possible to get `sourceMapOutputPath` in Gradle task as well?

I'm thinking that instead of creating a lone `.backtrace-sourcemap-id` file, you could save it to `${sourceMapOutputPath}.bt-debug-id` or something like that. That way, it'll be more connected to the file.

packages/react-native/processSourceMap.js
```diff
@@ -0,0 +1,62 @@
...
+
+    const debugIdGenerator = new DebugIdGenerator();
+    const sourceProcessor = new SourceProcessor(debugIdGenerator);
+    const { source, debugId } = sourceProcessor.processSource(bundle.code);
+    const snippet = debugIdGenerator.generateSourceSnippet(debugId);
+
+    console.debug(`Backtrace: saving debugId ${debugId} to ${backtraceSourceMapId}.`);
+    fs.writeFileSync(backtraceSourceMapId, debugId);
```
perf2711, line 37: small nit: this can be an async function, you could use promise-based functions here

examples/sdk/reactNative/ios-sourcemap-upload.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# Script responsible for preprocessing source maps with debugid and uploading it to Backtrace via backtrace-js. 
+# Usage:    ./ios-sourcemap-upload.sh <source_map_file_path> <debug_id_file_path> <backtrace_configuration_path
```
perf2711, line 4:
```suggestion
# Usage:    ./ios-sourcemap-upload.sh <source_map_file_path> <debug_id_file_path> <backtrace_configuration_path>
```

perf2711 (commented):
Can you add a description of setting this up in the React Native SDK README?

examples/sdk/reactNative/android/app/build.gradle
konraddysput: I left it uncommented because our example app should always show all benefits from our SDK. I think we want this option to be enabled by default.

examples/sdk/reactNative/ios-sourcemap-upload.sh
konraddysput: I'm assuming something might go wrong in an unknown environment. I still want to give the user and option to override the place where we generate a debugId.

packages/react-native/processSourceMap.js
konraddysput: async won't give us anything in this case. In my opinion this can cause additional headache and unexpected issues. This is a simple, fast and small operation. I would prefer to stay in the sync context and don't risk on executing a different part of the code faster than we wrote a file.

examples/sdk/reactNative/ios-sourcemap-upload.sh
perf2711: how do they override it? They would have to edit the script directly

packages/react-native/processSourceMap.js
perf2711: `customSerializer` is awaited in Metro, so there's no chance of that happening. I don't see any risk here.

examples/sdk/reactNative/ios-sourcemap-upload.sh
konraddysput: no - they need to set it in the env variable

packages/react-native/processSourceMap.js
konraddysput: sourceMapOutputPath might change depending on what is being invoked - sourceMapOutputPath is different in the metro parameter, and might be different later in gradle where the user override might be respected.

packages/react-native/processSourceMap.js
konraddysput: but at the same time - whats the point, what benefits it brings - the build system won't move, we need to block it anyway

examples/sdk/reactNative/README.md
```diff
@@ -8,53 +8,61 @@ This example app shows features available in the @backtrace/react-native package
 2. `npm install`. If you're on iOS, navigate to the `ios` directory and run `pod install`
 3. `npm run start` and pick desired platform
 
-#### Source maps
+### Source maps
+
+The example application is already integrated with the source map support. Once you change the .backtracejsrc file, source maps should be automatically uploaded to your project.
```
rick-bt, line 13:
```suggestion
This example application is integrated with the source map support. Once you change the .backtracejsrc file, source maps will be automatically uploaded to your project.
```

examples/sdk/reactNative/README.md
```diff
@@ -8,53 +8,61 @@ This example app shows features available in the @backtrace/react-native package
...
 
 Before executing any step:
 
--   Please update .backtracejsrc file with your symbols submission URL and your sourcemap settings.
+> Please update .backtracejsrc file with your symbols submission URL and your sourcemap settings.
 
-On Android: You can verify our example app with the source map support. In order to do that, please use the
-android-sourcemap.sh script.
+Backtrace is compatible with metro build system. In order to enable source map support, set a `customSerializer` method in the `metro.config.js` file to a function `processSourceMap` available in the `@backtrace/react-native/processSourceMap`.
```
rick-bt, line 19:
```suggestion
Backtrace is compatible with metro build system. To enable source map support, set a `customSerializer` method in the `metro.config.js` file to the `processSourceMap` function available in `@backtrace/react-native/processSourceMap`.
```

examples/sdk/reactNative/README.md
```diff
@@ -8,53 +8,61 @@ This example app shows features available in the @backtrace/react-native package
...
+module.exports = mergeConfig(getDefaultConfig(__dirname), config);
 
-On iOS: Backtrace simplify the flow needed to upload source maps by instructing hermesc how to generate and prepare
-source maps for your application. Without this additional step, react-native will not instruct your application to
-support Backtrace source maps.
+```
 
-In order to prepare your application for source maps and automatically upload them to Backtrace, modify your "Build
-Phase" with the following code:
+By adding Backtrace to the build chain, every time, your application is being prepared for the release builds, the Backtrace source map support will be automatically included.
```
rick-bt, line 34: Maybe
```suggestion
Add Backtrace to build automation to ensure every build has source map support.
```

examples/sdk/reactNative/README.md
```diff
@@ -8,53 +8,61 @@ This example app shows features available in the @backtrace/react-native package
...
-```bash
-set -e
+In order to upload source maps to Backtrace, you can:
+
+**On Android:**
 
-# destination source map directory
-SOURCE_MAP_DIR="$(pwd)/../build"
-mkdir -p $SOURCE_MAP_DIR
+In the `app/build.gradle` enable source map support by uncommenting hermes source map flags. With additional parameters, source maps will be generated. To automatically upload them to Backtrace, you can use the gradle task available the @backtrace/react-native library.
```
rick-bt, line 40:
```suggestion
Enable source map support in `app/build.gradle` by uncommenting hermes source map flags. With additional parameters, source maps will be generated. To automatically upload them to Backtrace, you can use the gradle task available the @backtrace/react-native library.
```

examples/sdk/reactNative/ios/reactNative.xcodeproj/project.pbxproj
```diff
@@ -269,7 +269,7 @@
...
+			shellScript = "set -e\n# enable source map support\nexport SOURCEMAP_FILE=\"$(pwd)/../main.jsbundle.map\"\n\nWITH_ENVIRONMENT=\"$REACT_NATIVE_PATH/scripts/xcode/with-environment.sh\"\nREACT_NATIVE_XCODE=\"$REACT_NATIVE_PATH/scripts/react-native-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n\n# upload source maps to Backtrace\nsource_map_upload=\"$(pwd)/../ios-sourcemap-upload.sh\"\nbacktrace_js_config=\"$(pwd)/../.backtracejsrc\"\n/bin/sh -c \"$source_map_upload $SOURCEMAP_FILE $TARGET_BUILD_DIR/.backtrace-sourcemap-id $backtrace_js_config\" \n\n";
```
melekr, line 272: is it possible to move the build phase script and upload script to the ios package like android, or does it complicate things with no real gain?

melekr, line 272: ios-sourcemaps.sh does a good job validating paths.
Build phase scripts sometimes fail silently, would be nice to add:

`if [ ! -f "$SOURCEMAP_FILE" ]; then
    echo "Error: Source map file not found at $SOURCEMAP_FILE"
    exit 1
fi`

`if [ -z "$REACT_NATIVE_PATH" ]; then
    echo "Error: REACT_NATIVE_PATH not set."
    exit 1
fi`

etc..

rick-bt (approved):

examples/sdk/reactNative/ios/reactNative.xcodeproj/project.pbxproj
konraddysput: We're following what react-native does. By doing that, we're always the last step in the build process. If you have an idea how we can move it to ios package - I'm happy to explore it with you.

konraddysput: SOURCEMAP_FILE will be created by the react-native build process. REACT_NATIVE_PATH is also set in a separated command just before this step. Our goal was to add minimum code to the existing available react-native build system to upload source maps. This is how it works in react-native now

packages/react-native/android/upload-sourcemaps.gradle
konraddysput: Two or three source maps are generated during the source map generation process. The last one is the one created by merging two other source maps (one from the metro build and second from hermesc for example). This is what is the final source map. Do you have a better idea for the name?

perf2711: I guess we could be a bit more verbose here. What do you think about this?

```suggestion
            println("Backtrace: Cannot find the final merged sourcemap from Metro and Hermes.")
```

packages/react-native/android/upload-sourcemaps.gradle
perf2711: What I mean is that you can omit the formatting, i.e. pretty string would be this:
```
{
  "a": 123
}
```
while non-pretty (what we want to save size):
```
{"a": 123}
```

packages/react-native/processSourceMap.js
perf2711: :+1:

packages/react-native/processSourceMap.js
perf2711: I guess none, just a matter of styling preference.

packages/react-native/android/upload-sourcemaps.gradle
konraddysput: It's changed to a default serialization option

perf2711 (approved):

melekr (approved):

PR #335, opened 2025-01-29
sdk-core, browser, node: add BacktraceApi

perf2711:
This change adds `BacktraceApi` class to `@backtrace/node` and `@backtrace/browser` packages that allows to directly interface with Backtrace submission API for reports and metrics.

Furthermore, HTTP request bodies and responses are added to `@backtrace/sdk-core`.

This change helps submitting reports and/or metrics to Backtrace without requiring to initialize the SDK. The SDK may be still initialized on its own.


packages/browser/src/BacktraceBrowserRequestHandler.ts
```diff
@@ -18,14 +22,8 @@ export class BacktraceBrowserRequestHandler implements BacktraceRequestHandler {
...
     };
-    constructor(
-        private readonly _options: {
-            url: string;
-            token?: string;
-            timeout?: number;
-        },
-    ) {
-        this._timeout = this._options.timeout ?? DEFAULT_TIMEOUT;
+    constructor(private readonly _options?: BacktraceBrowserRequestHandlerOptions) {
```
konraddysput, line 25: small nit: there is no need to keep _options in the class instance. we can remove `private readonly _` since we don't use it anywhere

packages/node/src/BacktraceNodeRequestHandler.ts
```diff
@@ -24,15 +30,8 @@ export class BacktraceNodeRequestHandler implements BacktraceRequestHandler {
...
-        private readonly _options: {
-            url: string;
-            token?: string;
-            timeout?: number;
-            ignoreSslCertificate?: boolean;
-        },
-    ) {
-        this._timeout = this._options.timeout ?? DEFAULT_TIMEOUT;
+    constructor(private readonly _options?: BacktraceNodeRequestHandlerOptions) {
+        this._timeout = this._options?.timeout ?? DEFAULT_TIMEOUT;
```
konraddysput, line 34: small nit: 

maybe we should destruct options and assign timeout/ignoreSslCertificate variables instead of keeping the _options object

packages/sdk-core/src/BacktraceCoreApi.ts
```diff
@@ -0,0 +1,97 @@
...
+            options.token,
+        );
+
+        this._uniqueMetricsSubmissionUrl = MetricsUrlInformation.generateUniqueEventsUrl(
+            options.metrics?.url ?? 'https://events.backtrace.io',
+            options.url,
+            options.token,
+        );
+
+        this._requestHandler = options.requestHandler ?? requestHandler;
```
konraddysput, line 48: requestHandler is always defined. It's weird that you want to override it - that's why you pass it as a second parameter, and you still use the one from the options. Maybe the client's backtrace API should override options if the request handler is not defined?

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitMemory.ts
```diff
@@ -0,0 +1,48 @@
+export interface DataBacktraceSubmitMemory {
```
konraddysput, line 1: how about renaming it to BacktraceDataSubmitMemory ? Sounds more natural and even the file name suggest it should start this way

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitSourceCode.ts
```diff
@@ -0,0 +1,61 @@
+export interface TextBacktraceSubmitSourceCode {
```
konraddysput, line 1: How about `BacktraceSubmitSourceCodeData` instead? In the UI source code is used for multiple purposes. We can check it with the UI team, but I think more generic name makes sense

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitStackFrame.ts
```diff
@@ -0,0 +1,66 @@
...
+    registers?: Record<string, number | string>;
+}
+
+export interface FuncBacktraceSubmitStackFrame extends BaseBacktraceSubmitStackFrame {
+    funcName: string;
+}
+
+export interface AddressBacktraceSubmitStackFrame extends BaseBacktraceSubmitStackFrame {
+    address: string;
+}
```
konraddysput, line 64: you can have both - for example: once you preprocess the frame you have funcName and address. I wouldn't use one or another.

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitStackFrame.ts
```diff
@@ -0,0 +1,66 @@
+interface BaseBacktraceSubmitStackFrame {
```
konraddysput, line 1: I really don't like the name. Can we rename it to the BacktraceSubmitStackFrameBase instead?

konraddysput (changes requested):
Can we work with names more?

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitStackFrame.ts
perf2711: You can have both - it's more that `funcName` or `address` is required. If you specify one, the other is not required then.

konraddysput (approved):

packages/sdk-core/src/BacktraceCoreApi.ts
perf2711: Removed requestHandler from core options.

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitMemory.ts
perf2711: renamed to `BacktraceSubmitDataMemory`

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitSourceCode.ts
perf2711: renamed to `BacktraceSubmitTextSourceCode`, I think the `text` part should stay, as it is describing what is actually in the model

packages/sdk-core/src/model/http/model/submit/request/BacktraceSubmitStackFrame.ts
perf2711: renamed to `BacktraceSubmitBaseStackFrame` to follow previous suggestions. `base` is an adjective here, not a noun, so I think it shouldn't be at the end.

PR #340, opened 2025-06-23
aibrahim/BT-5870/add-session-replay-to-browser-sdk-docs

GT1990:


packages/react-native/README.md
```diff
@@ -16,6 +16,7 @@ and easy, after which you can explore the rich set of Backtrace features.
     - [Attributes](#attributes)
     - [File Attachments](#file-attachments)
     - [Breadcrumbs](#breadcrumbs)
+    - [Session Replay](#session-replay)
```
rick-bt, line 19: Are you positive this works in React Native? I mean, I don't know that it doesn't, I would just like to be sure.

rick-bt (approved):

packages/react-native/README.md
GT1990: my mistake, only for sdks derived from browser sdk. removed react-native readme changes

rick-bt (approved):

PR #341, opened 2025-06-27
sdk-core: overhaul of database and storage [INT-355]

perf2711:
Overhauls storage and database, removing things such as file attachments from `sdk-core`, and allows them to be added in packages that actually can implement them.

Notable changes:
* add `BacktraceStorageModule` - behaves similarly to `FileSystem`, but omits directories and is more of a key/value storage
  * we do not use directories at all in `sdk-core`, so this is simpler and allows for easier implementations for storages like `localStorage` in browser
* allow for defining custom database records
  * removes file attachments from `sdk-core`, and by extent, `browser` and other SDKs that shouldn't care about files
  * allows to specify implementations for serialization and sending of database records


perf2711:
Do not expect the pipelines to succeed in this branch - this covers only `sdk-core`, and deriving packages will not build.

packages/electron/src/main/modules/BacktraceMainElectronModule.ts
```diff
@@ -149,7 +149,9 @@ export class BacktraceMainElectronModule implements BacktraceModule {
     ) {
```
konraddysput, line 149: there is a change from node/electron etc, is this expected?

perf2711: GitHub is drunk? The commits are on the `feature/INT-355-storage-overhaul` branch, they shouldn't be here...

perf2711: Yeah, I updated the base branch to the same branch and it's not here anymore.

konraddysput (approved):

PR #347, opened 2025-06-27
sdk-core, electron: fix endless loop when loading SessionFiles

perf2711:
When two session file markers have the same ID, but different timestamp, it results in an endless loop.

This change prevents that by also using the timestamp as session ID. 

This requires changes in Electron.

### DIFF
--- a/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
+++ b/packages/sdk-core/src/modules/database/BacktraceDatabase.ts
@@ -396,7 +396,7 @@ export class BacktraceDatabase extends Events<BacktraceDatabaseEvents> implement
         }
 
         const sessionId = record.sessionId;
-        if (typeof sessionId !== 'string') {
+        if (!SessionFiles.isValidSessionId(sessionId)) {
             this._sessionFiles.lockPreviousSessions(record.id);
             return;
         }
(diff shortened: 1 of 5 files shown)


packages/sdk-core/src/modules/storage/SessionFiles.ts
```diff
@@ -5,15 +5,19 @@ import { FileSystem } from './FileSystem.js';
 
 interface FileSession {
     readonly file: string;
-    readonly sessionId: string;
+    readonly sessionId: SessionId;
```
konraddysput, line 8: how about backward compatibility?

konraddysput (commented):
If someone already has session files - where do you handle a conversion from the old data model to the new one?

packages/sdk-core/src/modules/storage/SessionFiles.ts
perf2711: You're right, this could fail. I added a check for this where required.

packages/sdk-core/src/modules/database/BacktraceDatabase.ts
```diff
@@ -135,16 +135,14 @@ export class BacktraceDatabase extends Events<BacktraceDatabaseEvents> implement
...
-
         const record: ReportBacktraceDatabaseRecord = {
             type: 'report',
             data: backtraceData,
             timestamp: TimeHelper.now(),
             id: IdGenerator.uuid(),
             locked: false,
             attachments: attachments,
-            sessionId: typeof sessionId === 'string' ? sessionId : undefined,
+            sessionId: this._sessionFiles?.sessionId,
```
konraddysput, line 145: idea: you know when the file was created. Why not use a file creation date and drop the deprecated model at all?

konraddysput (approved):

PR #348, opened 2025-06-27
browser: add browser file system based on localStorage

perf2711:
A simple implementation of `BrowserFileSystem` based on `localStorage` that enables usage of `BacktraceDatabase` in the browser.

Requires https://github.com/backtrace-labs/backtrace-javascript/pull/347, as browsers have a single session ID and this causes endless loops.

### DIFF
--- a/packages/browser/src/storage/BrowserFileSystem.ts
+++ b/packages/browser/src/storage/BrowserFileSystem.ts
@@ -1,0 +1,90 @@
+import { BacktraceAttachment, BacktraceStringAttachment, FileSystem } from '@backtrace/sdk-core';
+
+const PREFIX = 'backtrace__';
+
+export class BrowserFileSystem implements FileSystem {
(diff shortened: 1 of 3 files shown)


packages/browser/src/storage/BrowserFileSystem.ts
```diff
@@ -0,0 +1,90 @@
+import { BacktraceAttachment, BacktraceStringAttachment, FileSystem } from '@backtrace/sdk-core';
+
+const PREFIX = 'backtrace__';
+
+export class BrowserFileSystem implements FileSystem {
```
konraddysput, line 5: This suppose to be an example, however we're including it in the library code. Are you sure about it?

konraddysput (approved):

GT1990 (approved):

konraddysput (approved):

PR #356, opened 2025-10-08
node: fix machine id not being a UUID in some cases

perf2711:
In some cases, like on Linux, the returned machine ID may not be a GUID, but a hostname:

```sh
( cat /var/lib/dbus/machine-id /etc/machine-id 2> /dev/null || hostname ) | head -n 1 || :
```

This change makes sure that non-GUID results are hashed and returned as GUIDs.


konraddysput (approved):

melekr (commented):
@konraddysput what's the status on this, good to merge?

melekr (commented):
@KishanPRao when we have time let's see if we can merge these.

PR #358, opened 2026-01-12
Rollup and vite upgrade

adamcronin42:
Updating the rollup and vite plugins to be compatible with the latest major versions as well as added an example in /examples/plugin/vite


examples/plugin/vite/package.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "dependencies": {
+    "@backtrace/browser": "^0.6.0"
+  },
+  "devDependencies": {
+    "@backtrace/vite-plugin": "file:../../../tools/vite-plugin",
+    "typescript": "^5.0.4",
+    "vite": "7.1.5"
+  }
+}
```
konraddysput: end of file character - can we add an enter?

konraddysput:
Shouldn't we add this an example to the example repo?

konraddysput (approved):

examples/plugin/vite/package.json
adamcronin42: Done

adamcronin42:
@konraddysput we didn't have any examples for the plugins, only sdks, so I created /examples/plugin. I can move it under /examples/sdk/ if you'd prefer though

konraddysput:
> @konraddysput we didn't have any examples for the plugins, only sdks, so I created /examples/plugin. I can move it under /examples/sdk/ if you'd prefer though

It does the same what our example do so maybe its just better to use an sdk example? Its up to you guys- its just an suggestion.

GT1990 (approved):

adamcronin42:
I feel you. I do like that it's an example that's vite specific though. I'll probably keep it if you don't mind.

PR #360, opened 2026-04-17
sdk-core: Serialize error.cause

konraddysput:
**Why**
This pull request allows for serializing errors.cause. Currently, if the cause exists, it's not visible in the annotation

### DIFF
--- a/packages/sdk-core/src/model/report/BacktraceReport.ts
+++ b/packages/sdk-core/src/model/report/BacktraceReport.ts
@@ -120,7 +120,7 @@ export class BacktraceReport {
                 error.cause instanceof Error && !seen.has(error.cause)
                     ? this.unwrapErrorToAnnotation(error.cause, seen)
                     : error.cause != null
-                      ? { value: String(error.cause) }
+                      ? { ...error.cause }
                       : undefined,
         };
     }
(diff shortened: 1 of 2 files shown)


perf2711 (approved):

melekr (approved):

packages/sdk-core/src/model/report/BacktraceReport.ts
```diff
@@ -111,6 +106,29 @@ export class BacktraceReport {
         }
     }
 
+    private unwrapErrorToAnnotation(
```
melekr: wondering if the current serialization is still dropping cause values 🤔

would something like this make sense ?

```suggestion
    private unwrapErrorToAnnotation(
    error: Error & { cause?: unknown },
    seen = new WeakSet<object>(),
): Record<string, unknown> {
    seen.add(error);

    const hasCause = 'cause' in error;
    const cause = error.cause;

    return {
        ...error,
        message: this.generateErrorMessage(error.message),
        name: error.name,
        stack: error.stack,
        cause:
            !hasCause
                ? undefined
                : cause instanceof Error
                  ? seen.has(cause)
                    ? `[Circular] ${cause.message}`
                    : this.unwrapErrorToAnnotation(cause, seen)
                  : cause,
    };
}
```

konraddysput: cool idea - yeah, it can simply the code. Applied!

PR #365, opened 2026-04-20
sdk-core: Allow to serialize and capture invalid attribute types

konraddysput:
**Why**

We notice, unavailable objects, or ones that can throw an error during toJSON methods can cause an SDK to generate an error. This pull request adds support for other type of objects in annotations/breadcrumbs

### DIFF
--- a/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
+++ b/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
@@ -145,11 +145,12 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
                         result[key] = value;
                         break;
                     }
+                    const unknownValue = value as unknown;
                     try {
-                        if (value instanceof Date) {
-                            result[key] = value.toISOString();
-                        } else if (value instanceof URL) {
-                            result[key] = value.toString();
+                        if (unknownValue instanceof Date) {
+                            result[key] = unknownValue.toISOString();
+                        } else if (unknownValue instanceof URL) {
+                            result[key] = unknownValue.toString();
                         }
                     } catch {
                         // revoked proxy or broken object — drop it
--- a/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
+++ b/packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
@@ -1,4 +1,5 @@
 import {
+    AttributeType,
     BacktraceAttachment,
     BacktraceAttachmentProvider,
     Breadcrumb,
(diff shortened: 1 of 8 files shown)


perf2711 (approved):

packages/sdk-core/src/common/limitObjectDepth.ts
```diff
@@ -2,29 +2,48 @@ type DeepPartial<T extends object> = Partial<{ [K in keyof T]: T[K] extends obje
 
 const REMOVED_PLACEHOLDER = '<removed>';
 
-export type Limited<T extends object> = DeepPartial<T> | typeof REMOVED_PLACEHOLDER;
+export type Limited<T> = (T extends object ? DeepPartial<T> : T) | typeof REMOVED_PLACEHOLDER;
+
+export function limitObjectDepth<T>(val: T, depth: number): Limited<T> {
```
melekr, line 7: the new `toJSON` path in `limitObjectDepth()` can recurse forever, a guard here like `serialized === val` + a regression test for a self-returning `toJSON()` case might be a good idea. 

Everything else lgtm.

melekr (approved):

packages/sdk-core/src/common/limitObjectDepth.ts
konraddysput: the depth parameter is responsible for handling it, do you think there is case it wont be respected?

melekr: edge case only under `toJSON()`, if it returns `this` or returns another object that leads back to the same object / another weird value.

please ignore my comment if we're sure :point_up: can't happen.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
```diff
@@ -107,9 +108,8 @@ export class FileBreadcrumbsStorage implements BreadcrumbsStorage {
             timestamp: TimeHelper.now(),
             type: BreadcrumbType[rawBreadcrumb.type].toLowerCase(),
             level: BreadcrumbLogLevel[rawBreadcrumb.level].toLowerCase(),
-            attributes: rawBreadcrumb.attributes,
+            attributes: this.prepareAttributes(rawBreadcrumb.attributes),
```
GT1990, line 111: @konraddysput InMemoryBreadcrumbsStorage.add still stores attributes untouched, so the two backends now are different. File drops nested objects and in-memory keeps them. Should we lift this into BreadcrumbsManager.add so both go through the same sanitizer?

packages/sdk-core/src/common/limitObjectDepth.ts
GT1990: @konraddysput I agree with @melekr the new toJSON path doesn't decrement depth when it recurses, so if toJSON() returns this (or anything that cycles back to an object with toJSON) we recurse forever and we'll get a stack overflow. Edge case, but it's the exact crash class this PR is guarding against.

packages/node/src/breadcrumbs/FileBreadcrumbsStorage.ts
konraddysput: Addressed in https://github.com/backtrace-labs/backtrace-javascript/pull/368

PR #368, opened 2026-04-20
sdk-core, node: In memory proper attribute filtering

konraddysput:
**Why**

Per @GT1990 comment, InMemoryAttributes are not covered. Therefore, I decided to move the implementation to the BreacrumbManager and always pass prepared attributes to the storage.

### DIFF
--- a/packages/sdk-core/src/modules/breadcrumbs/BreadcrumbsManager.ts
+++ b/packages/sdk-core/src/modules/breadcrumbs/BreadcrumbsManager.ts
@@ -185,6 +185,13 @@ export class BreadcrumbsManager implements BacktraceBreadcrumbs, BacktraceModule
             rawBreadcrumb.message = rawBreadcrumb.message.substring(0, this._limits.maximumBreadcrumbMessageLength);
         }
 
+        if (rawBreadcrumb.attributes) {
+            rawBreadcrumb = {
+                ...rawBreadcrumb,
+                attributes: this.prepareAttributes(rawBreadcrumb.attributes),
+            };
+        }
+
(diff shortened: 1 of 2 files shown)


GT1990 (approved):

packages/sdk-core/src/modules/breadcrumbs/BreadcrumbsManager.ts
```diff
@@ -207,6 +214,43 @@ export class BreadcrumbsManager implements BacktraceBreadcrumbs, BacktraceModule
...
+                    const unknownValue = value as unknown;
+                    try {
+                        if (unknownValue instanceof Date) {
+                            result[key] = unknownValue.toISOString();
+                        } else if (unknownValue instanceof URL) {
+                            result[key] = unknownValue.toString();
+                        }
+                    } catch {
+                        // revoked proxy or broken object — drop it
+                    }
```
perf2711, line 245: If you would extract `switch` to a separate method, you could `try/catch` for `result[key] = prepareSingleAttribute(unknownValue.toJSON());`, if `toJSON` method is available. `URL` and `Date` return `toString()` and `toISOString()` respectively under `toJSON`, so you would have them handled, and potentially other things.

Be sure to limit the depth in case of `toJSON` returning a value that is the same object, or also a value that has a `toJSON` function.

konraddysput: limitObjectDepth should be enough. I removed this change because the problem was existing somewhere else.

melekr (approved):

PR #369, opened 2026-04-21
sdk-core: Limit object depth

konraddysput:
**Why**

@melekr pointed out that a potential infinity loop can happen because of not decrementing object depth during recursion #365. This pull request properly decrease object depth


melekr:
awesome, TY!

melekr (approved):

PR #379, opened 2026-07-30
fix(react-native): report correct agent version and fix RN 0.72 init crash

KishanPRao:
## Summary
Reports carried a hardcoded `0.0.1` agent version, RN 0.72 apps crashed at SDK init because Metro 0.76.9 mis-indexes a babel helper require in sdk-core's bundle, and the sourcemap scripts never shipped, so JS symbolication could not be set up.

## Changes
* `packages/react-native/src/BacktraceClient.ts`: read `agentVersion` from the package version instead of the hardcoded `0.0.1`.
* `packages/sdk-core/tsconfig.build.json`: target ES5 with downlevelIteration so the shipped bundle has no class syntax for the consumer's Metro to transpile into babel helper requires.
* `packages/sdk-core/rollup.config.mjs`: bundle tslib so the ES5 output is self-contained.
* `packages/react-native/package.json`: add `scripts` to `files` so the sourcemap tooling ships.

## Note
- `agentVersion` reads `package.json` through the `react-native`/`src` entry Metro uses, so it is correct at runtime; the `lib/` builds resolve the path one level too deep, but RN never loads `lib/`.
- RN 0.72 on Xcode 26 also needs an app-side Podfile snippet for Yoga's `-Werror` (client side, not SDK); ES5 costs sdk-core ~8KB gzip across all JS SDKs.
- Validated RN 0.72-0.83 (both architectures); unit tests across sdk-core/react-native/browser/react/node.
- `test_linux_all` is red due to a Sauce Connect service-side issue, unrelated to this change. Instead, validated on Chrome, Firefox, Edge, Safari, manually.

ref: BT-7354, BT-7406

### DIFF
--- a/.github/actions/smoke-tests/action.yml
+++ b/.github/actions/smoke-tests/action.yml
@@ -34,9 +34,13 @@ runs:
                       - 'packages/browser/**'
                       - 'packages/react/**'
                       - 'packages/sdk-core/**'
+                      - 'smoketests/**'
+                      - '.github/actions/smoke-tests/**'
                   node:
                       - 'packages/sdk-core/**'
                       - 'packages/node/**'
+                      - 'smoketests/**'
+                      - '.github/actions/smoke-tests/**'
 
         - run: npm run smoketest:node
           shell: bash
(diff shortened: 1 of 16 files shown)


melekr (commented):
we're pinning "react-native": ">=0.72.0"

melekr (approved):

PR #380, opened 2026-07-30
fix(react-native): skip duplicate native report for fatal unhandled JS errors on iOS

KishanPRao:
## Summary
On iOS a fatal unhandled JS error was reported twice (a JS "Unhandled exception" plus a redundant native "Crash"); this drops the native duplicate while keeping the JS report.
Adds a jest test for the marker, and makes the example's "Throw an unhandled error" button a real fatal.

## Changes
* `UnhandledExceptionHandler.ts`: on a fatal error, call `CrashReporter.markFatalError()` before chaining the previous global handler.
* `CrashReporter.ts`: add an optional-chained static `markFatalError()`, a no-op where the native binding is absent (Android).
* `BacktraceReactNative.mm`: export `markFatalError` as a blocking-synchronous method so the flag is set before `RCTFatal`.
* `BacktraceCrashReporter.h`: declare `markJsFatalError`.
* `BacktraceCrashReporter.mm`: stash a `jsFatal` flag in the crash's PLCrashReporter `customData`, purge that pending report on next launch, self-clear the flag after 2s.
* `unhandledExceptionHandlerTests.spec.ts`: cover `markFatalError` firing on a fatal error and not on a non-fatal one.
* `actions.ts`: make "Throw an unhandled error" a real fatal via a deferred throw instead of an async rejection.

## Note
The 2s self-clear calls `saveReportData` on the main queue, which can race the module-thread attribute setters; harmless (a stale attribute at worst), can be serialized in a follow-up.

ref: BT-7403

### DIFF
--- a/examples/sdk/reactNative/src/actions/actions.ts
+++ b/examples/sdk/reactNative/src/actions/actions.ts
@@ -78,9 +78,9 @@ export function generateActions(client: BacktraceClient) {
         {
             name: 'Throw an unhandled error',
             platform,
-            action: async () => {
-                notify('Sending an unhandled exception to Backtrace.');
-                throwAnError();
+            action: () => {
+                // fatal synchronous throw (ErrorUtils -> RCTFatal on iOS), not an async promise rejection
+                setTimeout(() => throwAnError(), 0);
             },
         },
         {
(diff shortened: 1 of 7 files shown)


melekr (approved):
🚀

PR #388, opened 2026-08-20
chore(ci): add Android emulator job to CI

KishanPRao:
## Summary
CI now tests the vendored Android native libraries through an emulator job that captures a real native crash, and asserts the report carries attributes set before and after init.

## Changes
* `.github/workflows/android.yml`: emulator job on ubuntu with KVM, api 34 x86_64, matrixed over New and Old Architecture, 16KB alignment gate first, npm/gradle/AVD caching, paths filter.
* `.github/scripts/verify-jni-symbols.sh`: fails when an ABI is missing from the APK or its library lacks the JNI entry points the SDK binds.
* `.github/scripts/ci-crash-driver.js`: deep-link listener appended to the example at CI time, sets a per-run marker attribute and crashes, logs an armed line.
* `.github/scripts/android-native-crash.sh`: waits for the armed handshake, fires the deep link, pulls the minidump binary-safe with a size-stability check.
* `.github/scripts/check-minidump-annotations.py`: asserts the marker key and value plus an init-time attribute as length-prefixed annotations, not raw strings.
* `examples/sdk/reactNative/android/app/build.gradle`: opt-in `btCiDebuggable` so CI can read the crashpad database on a release build.

ref: BT-7515


melekr (commented):
awesome

PR #390, opened 2026-08-21
chore(ci): restore sauce browser smoke tests

KishanPRao:
## Summary
Restores the sauce browser smoke tests, broken since the SC4 cutoff.

## Changes
* `.github/workflows/test.yml`: run `smoke_browser_sauce` on PRs again, pre-install the sc binary.
* `.github/actions/smoke-tests/action.yml`: run smoke tests when workflow files change.
* `smoketests/wdio.conf.ts`: pin sc to 5.5.3, proxy localhost directly.

ref: BT-7528


melekr (commented):
send it !

PR #391, opened 2026-08-26
feat(react-native): detect and report Android ANRs

KishanPRao:
## Summary

Adds Android ANR detection with two modes, a threshold watchdog that reports mid-hang and ApplicationExitInfo records reported on the next launch, both carrying every thread's stack.

## Changes

**`packages/react-native/src`**

* `BacktraceConfiguration.ts`: adds the `BacktraceAnrType` enum and `anr` configuration options.
* `BacktraceClient.ts`: starts ANR reporting on initialize, picks watchdog or exit-info by type, disposes the handler.
* `anr/AnrWatchdogHandler.ts`: turns watchdog events into reports with main as the faulting thread and every other thread included, pauses the watchdog while backgrounded.
* `anr/AnrReporter.ts`: reads exit-info records on launch, sends oldest first, saves the last-timestamp cursor when a report is delivered or stored in the database for retry, attaches the OS dump.
* `anr/AnrThreads.ts`: attaches the non-main threads to a report, renaming duplicate thread names.
* `anr/AnrException.ts`: error named `Hang` for classification.
* `index.ts`: exports the new enum and config type.

**`packages/react-native/android`**

* `BacktraceAnrWatchdog.java`: native module, starts/stops the watchdog, emits detection events with frames for every thread.
* `BacktraceReactNative.java`: adds `getAnrExitInfo`, logs read failures at warn.
* `ReactNativePackage.java`: registers the watchdog module.
* `AnrExitInfoReader.java`: collects REASON_ANR records, returns them oldest first with attributes, message, raw trace, and parsed frames for the main thread and every other thread in the dump.
* `anr/AnrWatchdog.java`: watchdog thread, posts a ticker to the main looper, reports when the tick is missed past the timeout, one report per hang, optional debugger bail-out.
* `anr/BacktraceThreadWatcher.java`: tick counters the watchdog compares.
* `anr/ExitInfo.java`: ported from backtrace-android.
* `anr/ExitInfoAdapter.java`: ported from backtrace-android.
* `anr/ProcessExitInfoProvider.java`: ported from backtrace-android.
* `anr/ActivityManagerExitInfoProvider.java`: exit-record lookup, fetches all records instead of a fixed count.
* `anr/ExitInfoStackTraceParser.java`: parses the OS ANR dump into per-thread frames, keeps native-method, unknown-source, and fileless frames, null-guards unnamed threads.
* `anr/AppExitInfoDetailsExtractor.java`: extracts the ANR attributes and OS trace, attribute names match backtrace-android.
* `anr/StackFrameMapper.java`: maps stack traces to bridge frames and formatted text, keeps the native program counter as `address`.
* `build.gradle`: adds junit for unit tests.

**Tests**

* `tests/anrReporterTests.spec.ts`: covers ordering, cursor advancement, database-stored failures, attributes, attachment.
* `tests/anrWatchdogHandlerTests.spec.ts`: covers event handling, thread-name dedup, background pause, dispose.
* `android/src/test/ExitInfoStackTraceParserTest.java`: parser tests over a real ANR dump, covering the fileless and duplicate-name forms.
* `android/src/test/resources/anrAppExitInfoStacktrace.txt`: the dump fixture.

**Docs and CI**

* `README.md`: ANR Detection section.
* `.github/workflows/android.yml`: runs the SDK unit tests before the emulator build.

**`examples/sdk/reactNative`**

* `index.js`: enables ANR in the example.
* `metro.config.js`: pins a single react-native copy, a duplicate splits the JS event registry from the native one.
* `ErrorGenerator.java`: adds `blockMainThread` to trigger a real ANR.
* `src/actions/android/action.ts`: the ANR action blocks the main thread instead of sleeping async.

## Note

A threshold report can be lost if the OS kills the app before the hang clears; the applicationExit type survives the kill.

ref: BT-7451


packages/react-native/src/anr/AnrReporter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            // records arrive oldest first, so stopping on failure leaves the rest for the next launch
+            for (const record of records) {
+                if (!record.mainThreadFrames?.length) {
+                    // no groupable stack, but still mark it seen or it is re-read on every launch
+                    await this.writeMarker(record.timestamp);
+                    continue;
+                }
+
+                const result = await client.send(this.buildReport(record));
```
melekr, line 73: Could we distinguish between a report that still needs retrying and one that has already been accepted locally? Thinking a solution would be for send() to return an explicit stored result, with an implementation that could observe the database insertion associated with the submission.

packages/react-native/src/anr/AnrReporter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const applicationDirectory = fileSystem.applicationDirectory();
+        if (!applicationDirectory) {
+            return undefined;
+        }
+
+        return new AnrReporter(fileSystem, `${applicationDirectory}/${AnrReporter.MarkerFileName}`);
+    }
+
+    public async report(client: BacktraceClient): Promise<void> {
```
melekr, line 59: a timestamp only cursor might skip a failed record permanently here ~

packages/react-native/android/src/main/java/backtrace/library/anr/ExitInfoStackTraceParser.java
```diff
@@ -0,0 +1,219 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ExitInfoStackTraceParser {
+    private static final Pattern JAVA_FRAME_PATTERN = Pattern.compile("\\s*at (.*?)\\((.*?):(\\d+)\\)");
```
melekr: support Java frames that do not include file:line


```suggestion
   private static final Pattern JAVA_FRAME_PATTERN =
        Pattern.compile(
                "^\\s*at\\s+(.+?)\\("
                        + "(?:"
                        + "(.+?):(\\d+)"
                        + "|(Native [Mm]ethod)"
                        + "|(Unknown Source)"
                        + "|(.+?)"
                        + ")"
                        + "\\)\\s*$");

static StackTraceElement parseJavaFrame(String frame) {
    Matcher matcher = JAVA_FRAME_PATTERN.matcher(frame);
    if (!matcher.matches()) {
        return null;
    }

    String qualifiedMethod = matcher.group(1);
    int lastDot = qualifiedMethod.lastIndexOf('.');

    String className = lastDot < 0
            ? qualifiedMethod
            : qualifiedMethod.substring(0, lastDot);

    String methodName = lastDot < 0
            ? ""
            : qualifiedMethod.substring(lastDot + 1);

    if (matcher.group(4) != null) {
        // StackTraceElement uses -2 to represent a native method.
        return new StackTraceElement(
                className,
                methodName,
                null,
                -2);
    }

    if (matcher.group(5) != null) {
        return new StackTraceElement(
                className,
                methodName,
                null,
                -1);
    }

    if (matcher.group(6) != null) {
        return new StackTraceElement(
                className,
                methodName,
                matcher.group(6),
                -1);
    }

    return new StackTraceElement(
            className,
            methodName,
            matcher.group(2),
            Integer.parseInt(matcher.group(3)));
}
```

packages/react-native/android/src/main/java/backtrace/library/anr/StackFrameMapper.java
```diff
@@ -0,0 +1,55 @@
...
+        for (StackTraceElement frame : frames) {
+            formatted.append(frame.toString()).append('\n');
+        }
+        return formatted.toString();
+    }
+
+    private static WritableMap toWritableFrame(StackTraceElement frame) {
+        WritableMap map = new WritableNativeMap();
+
+        String fileName = frame.getFileName();
```
melekr, line 31: discarded parsed native program counter


```suggestion
        if (fileName != null
        && fileName.startsWith(NATIVE_FILE_PREFIX)) {
    map.putString(
            "funcName",
            frame.getMethodName());

    map.putString(
            "library",
            frame.getClassName());

    map.putString(
            "address",
            fileName.substring(
                    NATIVE_FILE_PREFIX.length()));

    return map;
}
```

melekr (approved):
🚀

packages/react-native/src/anr/AnrReporter.ts
KishanPRao: Good idea!
Moving the cursor for failed records now, so we avoid duplicate reports being sent.

packages/react-native/src/anr/AnrReporter.ts
KishanPRao: Yes, now we move cursor when the report is persisted into the database, since future retries are handled through it.

packages/react-native/android/src/main/java/backtrace/library/anr/StackFrameMapper.java
KishanPRao: Updated to include native pc.

packages/react-native/android/src/main/java/backtrace/library/anr/ExitInfoStackTraceParser.java
KishanPRao: Updated to include Java, and also match missing native, file only frame patterns.

PR #393, opened 2026-09-02
bugfix: deliver unhandled java exception reports before process death

KishanPRao:
## Summary
In the default configuration an unhandled Java exception report never left the device: the handler killed the process right after handing the exception to JS.

## Changes
* `BacktraceAndroidBackgroundUnhandledExceptionHandler.java`: blocks the crashing thread on a `CountDownLatch` until JS calls the new `reportProcessed` method, 5s cap, invokes the JS callback once, synchronizes `uncaughtException`.
* `AndroidUnhandledExceptionHandler.ts`: awaits `client.send`, calls `reportProcessed` from a finally, swallows a failed send.
* `androidUnhandledExceptionHandlerTests.spec.ts`: new, covers signalling native after a successful and a failed send, the error type, no start without the bridge, stop on dispose.

ref: BT-7513


melekr (approved):
🚀

PR #395, opened 2026-09-08
feature: ios sourcemap upload script fails debug builds and manual runs

KishanPRao:
## Summary
Debug iOS builds no longer fail on the Backtrace source map upload phase, and the upload script now runs outside Xcode.

## Changes
* `packages/react-native/scripts/ios-sourcemap-upload.sh`: skips Debug and `SKIP_BUNDLING` builds (dev bundles get no debug id), reports a missing source map, debug id, config, `backtrace-js` or node as an Xcode warning instead of a build failure, keeps them as errors outside Xcode, reads `NODE_BINARY` from `ios/.xcode.env` instead of `$REACT_NATIVE_PATH/scripts/node-binary.sh`.
* `packages/react-native/tests/scripts/iosSourcemapUploadScript.spec.ts`: new, covers Debug/Release/custom configurations, `SKIP_BUNDLING`, missing inputs, `.xcode.env` node resolution, `DEBUG_ID_PATH`, upload failure, and manual runs.

ref: BT-7405


melekr (approved):
🚀

